PR #9, opened 2025-05-20
Flavor Farm Corto Demo

therealaditigupta:
Modified the corto demo code to reflect the 1-1 supervisor <-> farm interaction where the supervisor provides a prompt that specifies a location and a season and the farm provides a crafty string with a flavor profile of what the coffee may taste like from the given location and season.

Includes:
- detailed Agent card
- environment variable handling via a config file. This ensures the supervisor and the farm agent are using the farm host and port set in environment variables with a default of localhost:9999
- keeps original langGraph from both sides with changes to the inputs and outputs 
- LLM prompt that parses the provided prompt to get the location and season and then gets the flavor profile for this combo
- A2A error handling using A2A recommended error types
- Detailed README for current Corto implementation (can change as we add more features/complexity)

### DIFF
--- a/FlyingCircusRoastery/coffee_agents/corto/config/config.py
+++ b/FlyingCircusRoastery/coffee_agents/corto/config/config.py
@@ -7,13 +7,4 @@ load_dotenv()  # Automatically loads from `.env` or `.env.local`
 FARM_AGENT_HOST = os.getenv("FARM_AGENT_HOST", "localhost")
 FARM_AGENT_PORT = int(os.getenv("FARM_AGENT_PORT", "9999"))
 
-# --- OpenAI ---
-OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")
-OPENAI_MODEL = os.getenv("OPENAI_MODEL", "gpt-4")
-OPENAI_TEMPERATURE = float(os.getenv("OPENAI_TEMPERATURE", "0.7"))
-
-# --- Azure OpenAI ---
-AZURE_OPENAI_API_KEY = os.getenv("AZURE_OPENAI_API_KEY")
-AZURE_OPENAI_ENDPOINT = os.getenv("AZURE_OPENAI_ENDPOINT")
-AZURE_OPENAI_API_VERSION = os.getenv("AZURE_OPENAI_API_VERSION")
-AZURE_OPENAI_DEPLOYMENT_NAME = os.getenv("AZURE_OPENAI_DEPLOYMENT_NAME")
+LLM_PROVIDER = os.getenv("LLM_PROVIDER")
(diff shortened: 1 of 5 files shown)


FlyingCircusRoastery/coffee_agents/corto/config/config.py
```diff
@@ -0,0 +1,19 @@
...
+from dotenv import load_dotenv
+
+load_dotenv()  # Automatically loads from `.env` or `.env.local`
+
+# --- Corto ---
+FARM_AGENT_HOST = os.getenv("FARM_AGENT_HOST", "localhost")
+FARM_AGENT_PORT = int(os.getenv("FARM_AGENT_PORT", "9999"))
+
+# --- OpenAI ---
+OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")
```
shanchunyang0919: Should we use a shared `LLMFactory` library (OSPO-approved and maintained by SRE) for this?  
Reference: [LLMFactory in jira_agent](https://github.com/cisco-outshift-ai-agents/jira-agntcy-agent/blob/main/jira_agent/common/llm.py#L19)

I recommend keeping environment variable names consistent across agents and repositories. For example, the names used by both Jarvis and Jira agents are:
```plaintext
AZURE_OPENAI_DEPLOYMENT: gpt-4o
AZURE_OPENAI_API_VERSION: <version>
AZURE_OPENAI_ENDPOINT: <endpoint>
AZURE_OPENAI_API_KEY: <api_key>
AZURE_OPENAI_TEMPERATURE: 0.7
OPENAI_ENDPOINT: <openai_endpoint>
OPENAI_API_KEY: <openai_api_key>
OPENAI_API_VERSION: gpt-4o
LLM_PROVIDER: openai
```

therealaditigupta: Absolutely, we can use the LLMFactory. Is there any other documentation on what LLMFactory does and how to use it?

I used the same variable names used by the Alfred and subsequent agents, but if LLMFactory has a standard, happy to follow it

shanchunyang0919: Here's an simple example of how it could be along with langchain `create_react_agent`
```
class ProjectsAgent:

    def __init__(self, settings=None):
        self.name="jira_projects_agent"
        self.tools=tools
        self.prompt=prompt.format(additional_context="")

    def agent(self):
        agent = create_react_agent(
            name=self.name,
            model=LLMFactory(
                provider=os.getenv("LLM_PROVIDER"),
            ).get_llm(),
            tools=self.tools,
            prompt=self.prompt,
            response_format=LLMResponseOutput
        )
        return agent
```

shanchunyang0919: I'll mark this discussion as resolved since we've come up with a solution.

therealaditigupta: This change has been implemented.

PR #10, opened 2025-05-20
Gateway sdk integration

codyhartsook:
Integrating the gateway sdk factory methods for clients and bridges.

Additions
- FarmServer is served using gateway bridge via AGP
- Exchange agent establishes an A2A client with AGP as transport
- A2A agent card based topic messaging instead of URL passing
- docker-compose with AGP dataplane server


FlyingCircusRoastery/coffee_agents/corto/config/docker/agp/server-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    pubsub:
+      servers:
+        - endpoint: "0.0.0.0:46357"
+          tls:
+            insecure: true
+
+      clients: []
+    controller:
+      server:
+        endpoint: "0.0.0.0:46358"
```
nandu-gh, line 25: Has AGP control plane been released? Even if it is released, why do we need it now?

FlyingCircusRoastery/coffee_agents/corto/exchange/graph/nodes.py
```diff
@@ -0,0 +1,66 @@
...
+class FarmContact:
+    """ 
+    A2A client for the Remote Coffee Farm Agent. 
+    """
+    def __init__(self, remote_agent_card: AgentCard):
+        self.remote_agent_card = remote_agent_card
+        self._a2a_topic = A2AProtocol.create_agent_topic(self.remote_agent_card)
+        
+        self._factory = GatewayFactory()
+        self._transport = self._factory.create_transport(DEFAULT_MESSAGE_TRANSPORT, AGP_GATEWAY_URL)
```
nandu-gh, line 24: So, are we not having an `Options` object encapsulating the specific transport, in this case that being `AGP`?

FlyingCircusRoastery/coffee_agents/corto/exchange/graph/nodes.py
```diff
@@ -0,0 +1,66 @@
...
+    async def invoke(self, state: dict):
+        resp = await self.message_farmer(state["input_payload"])
+        return {"output": resp}
+
+    async def message_farmer(self, input_payload: dict):
+        """ Sends a message to the agent server and returns the response. """
+        print(f"Sending message to {self.remote_agent_card.name} with payload: {input_payload}")
+
+        if not self._a2a_client:
+            self._a2a_client = await self._factory.create_client("A2A", agent_topic=self._a2a_topic, transport=self._transport)
```
nandu-gh, line 39: Looks like when the first message (prompt/ input request) comes, that's when the connection to the AGP data plane is established? If there is an issue with connection establishment, then the client would get an error. Can the connection be established with the AGP DP earlier?

FlyingCircusRoastery/coffee_agents/corto/farm/agent.py
```diff
@@ -1,26 +1,81 @@
...
+        season = state.get("season")
+        location = state.get("location")
+
+        if not season or not location:
+            return {"error": "Missing 'season' or 'location' in input."}
+
+        user_prompt = (
+            f"The coffee farm is located in {location}, and the current season is {season.capitalize()}. "
+            "Based on this information, what flavor profile can we expect from the coffee beans harvested here?\n\n"
+            "Please describe expected flavor notes using coffee-tasting terminology (e.g. acidity, body, aroma, finish), "
```
nandu-gh, line 32: What is the `temperature` value by default when the LLM call is made?

FlyingCircusRoastery/coffee_agents/corto/config/docker/agp/server-config.yaml
codyhartsook: There is a minimal control plane but we are not running it at the moment. This is copied directly from the AGP base server-config but we could remove the controller endpoint. I would like to explore using the control plane for service discovery in the next release.

FlyingCircusRoastery/coffee_agents/corto/exchange/graph/nodes.py
codyhartsook: At the factory level it is not necessarily clear how to pass options to the transport so I've adopted the approach of either passing an already instantiated client (AGP or nats) that we wrap with callbacks, or using our default options. This gives full controls or minimal effort as a tradeoff. Happy to hear suggestions.

At the interface level for the transport there are these two options for instantiation:

@classmethod
    @abstractmethod
    def from_client(cls: Type[T], client: Any) -> T:
        """Create a transport instance from a client."""
        pass

    @classmethod
    @abstractmethod
    def from_config(cls: Type[T], endpoint: str, **kwargs) -> T:
        """Create a transport instance from a configuration."""
        pass

FlyingCircusRoastery/coffee_agents/corto/exchange/graph/nodes.py
codyhartsook: Yes I can update this. The issue is that connecting is an async method and the agent is not async until graph invoke. I can try an asyncio.run(connect()) from the synchronous constructor

FlyingCircusRoastery/coffee_agents/corto/farm/agent.py
therealaditigupta: We aren't setting a temperature at the moment and OpenAI's default temperature is 1.0. We can set a value if needed

FlyingCircusRoastery/coffee_agents/corto/README.md
```diff
@@ -8,6 +8,12 @@ Ensure you have `uv` installed. You can install it using Homebrew:
 brew install uv
 ```
 ## Quick Start
+**Step 0: Deploy the message bus server**
+To run the messaging server (AGP | NATS | MQTT), bring up the services using Docker Compose.
+```sh
+docker-compose up
```
nandu-gh, line 14: There is no `docker-compose` tool anymore. `docker` CLI has a compose option.

FlyingCircusRoastery/coffee_agents/corto/README.md
```diff
@@ -8,6 +8,12 @@ Ensure you have `uv` installed. You can install it using Homebrew:
 brew install uv
 ```
 ## Quick Start
+**Step 0: Deploy the message bus server**
+To run the messaging server (AGP | NATS | MQTT), bring up the services using Docker Compose.
```
nandu-gh, line 12: Remove MQTT and don't bother with MQTT support for now even though code might support it.

FlyingCircusRoastery/coffee_agents/corto/config/config.py
```diff
@@ -0,0 +1,21 @@
+import os
+from dotenv import load_dotenv
+
+load_dotenv()  # Automatically loads from `.env` or `.env.local`
+
+# --- Corto ---
+DEFAULT_MESSAGE_TRANSPORT = os.getenv("DEFAULT_MESSAGE_TRANSPORT", "AGP")
+AGP_GATEWAY_URL = os.getenv("AGP_GATEWAY_URL", "http://localhost:46357")
+FARM_AGENT_HOST = os.getenv("FARM_AGENT_HOST", "localhost")
```
nandu-gh, line 9: Are these `FARM` env variables correct. Check with John.

FlyingCircusRoastery/coffee_agents/corto/config/docker/agp/server-config.yaml
nandu-gh: Ok.

FlyingCircusRoastery/coffee_agents/corto/exchange/graph/nodes.py
nandu-gh: Thanks for making the changes.

PR #20, opened 2025-05-29
feat: update A2A version to pypi a2a-sdk==0.2.5

codyhartsook:
This PR updates the pyproject.toml to import A2A from pypi with version a2a-sdk==0.2.5. The A2A exchange client and farm server have been updated to accommodate the version update. 

1. Update the farm server agent executor to implement new A2A AgentExecutor interface
2. Update the farm server request handler to use the updated agent executor
3. Update the farm server AgentCard with latest fields
4. Update the exchange A2A client send_message parameters

### DIFF
--- a/coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
+++ b/coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
@@ -5,8 +5,9 @@ from a2a.server.events import EventQueue
 from a2a.utils.errors import ServerError
 from a2a.types import (
     UnsupportedOperationError,
+    JSONRPCResponse,
+    ContentTypeNotSupportedError,
     InternalError,
-    InvalidParamsError,
     Task)
 
 from a2a.utils import (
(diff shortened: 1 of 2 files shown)


coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
```diff
@@ -1,91 +1,64 @@
...
     def __init__(self):
         self.agent = FarmAgent()
 
-    @override
-    async def on_message_send(
-            self,
-            request: SendMessageRequest,
-            event_queue: EventQueue,
-            task: Task | None,
+    def _validate_request(self, context: RequestContext) -> bool:
```
therealaditigupta: Is this required to be implemented by A2A? Just wondering why we need this?

coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
```diff
@@ -1,91 +1,64 @@
...
+        return False
+    
+    async def execute(
+        self,
+        context: RequestContext,
+        event_queue: EventQueue,
     ) -> None:
-        """
```
therealaditigupta, line 26: Why are we removing this doc string? ideally we can update it with the new parameters

coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
```diff
@@ -1,91 +1,64 @@
...
-        Handle incoming message requests to generate a flavor profile for coffee beans.
-        This method processes the request, invokes the FarmAgent to generate a flavor profile,
-        and enqueues the response message to the event queue.
-
-        Args:
-            request (SendMessageRequest): The incoming message request containing the agent message.
-            event_queue (EventQueue): The event queue to enqueue the response message.
-            task (Task | None): Optional task context, not used in this implementation.
-        """
-        logger.info("Received message request: %s", request)
```
therealaditigupta, line 36: Let's keep this logger line as well. Default logging level is set to info and this allows users running the demo to see the progression of both agents from the terminal.

coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
therealaditigupta: I see the reason for this function. Do we want to add any validation logic here to ensure there's user input? I see that we are fetching user input via prompt = context.get_user_input(). Does this handle the case where there is no user message passed in?

codyhartsook: I will look into adding validation logic on the input and review the get_user_input method. Langgraph and autogen samples both use get_user_input, I'll track down if those use multiple message types

coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
codyhartsook: May have deleted doc string on accident, will add back

coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
codyhartsook: Added the logging statement back

coffeeAGNTCY/coffee_agents/corto/farm/agent_executor.py
codyhartsook: get_user_input method maps to get_message_text which simply joins all text content from a Message's parts.

PR #21, opened 2025-05-29
fix: docker compose path, readme changes

Shridhar-2205:


coffeeAGNTCY/coffee_agents/corto/README.md
```diff
@@ -88,3 +88,50 @@ curl -X POST http://127.0.0.1:8000/agent/prompt \
 ```
 
 The `exchange` sends user inputs to the `farm_server` and displays the generated flavor profiles. It interacts with the `farm_server` through A2A communication protocols.
+
+
+## Docker Compose Setup
+
+**Prerequisite:**
+Before running Docker Compose, copy the example environment file and fill in your credentials:
```
therealaditigupta, line 133: IMO we don't need the environment variable steps again (copy the file, fill in the env variables) since they are already referenced in the setup section of the same README.md.

Shridhar-2205: Oh yeah, I thought that too based on your chat msg before, but what if we just want to use docker compose stack and avoid local  setup steps ? With UI ?

PR #71, opened 2025-07-09
README with Updated Picture Assets

therealaditigupta:


coffeeAGNTCY/coffee_agents/corto/farm/agent.py
```diff
@@ -76,7 +76,7 @@ async def flavor_node(self, state: State):
             SystemMessage(content=system_prompt),
             HumanMessage(content=user_prompt)
         ]
-        response = await get_llm().invoke(messages)
+        response = get_llm().invoke(messages)
```
Shridhar-2205, line 79: this should fix the issue right ?

therealaditigupta: Yes I think so!

Issue #96, opened 2025-07-31
[Bug]: corto-ui deployment with helm fails due to missing serviceaccount


MichaelClifford:
### Bug Description

When using helm to deploy the corto-ui example to a Kubernetes cluster, it fails with a `serviceaccount "corto-ui-sa" not found'` error. 

It looks like this error occurs due to the fact that in `values.yaml` the service account values are commented out by default. This causes the `serviceaccount.yaml` template to not get rendered, however, the `deployment.yaml` still renders an entry for the `serviceAccountName`. This leads to a state where the deployment is looking for a non-existent service account and fails. 

Including the `serviceAccountName` entry in the `deployment.yaml` should also be conditioned on the existence of service account values in `values.yaml`  

### Expected Behavior

The default values in the helm chart should lead to a successful deployment of the corto-ui example. 

### Affected Version

v1.0.0

### Steps to Reproduce

1. Install the helm chart:
```
helm install corto-ui coffeeAGNTCY/coffee_agents/corto/deployment/helm/corto-ui
```
2. Examine the `deployment.yaml`:
```
kubectl get deployment corto-ui -o yaml
```
3. See the error:
```
...
serviceaccount "corto-ui-sa" not found'
```

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

MichaelClifford:
The service account details are commented out in `values.yaml` and I believe that is what is causing the issue. https://github.com/agntcy/coffeeAgntcy/blob/f7381b71cd7d79638d9c80325641b73580172483/coffeeAGNTCY/coffee_agents/corto/deployment/helm/corto-ui/values.yaml#L20-L22 

With the service account details commented out, `serviceaccount.yaml` never gets created during `helm install`. However, the `deployment.yaml` still expects one to exists. 

https://github.com/agntcy/coffeeAgntcy/blob/f7381b71cd7d79638d9c80325641b73580172483/coffeeAGNTCY/coffee_agents/corto/deployment/helm/corto-ui/templates/deployment.tpl.yaml#L23-L26 

#97 tries to address this by making the need for a serviceAccountName conditioned on the existence of a service account. But maybe a better solution is to uncomment the service account details and provide some generic defaults in `values.yaml`? Let me know what you think.

PR #97, opened 2025-07-31
fix: corto-ui helm chart deployment template optional service account

MichaelClifford:
# Description

This PR addresses #96 

This PR adds an `if` statement around the `serviceAccountName` field in the `deployment.tpl.yaml` in the corto-ui example so that it is only rendered if there are values for the service account in `values.yaml` 

This is needed since the service account values are commented out by default, leading to a situation where the `deployment.yaml` defines a `serviceAccountName` without the corresponding service account being created. 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


Shridhar-2205 (commented):
we have sample in https://github.com/MichaelClifford/coffeeAgntcy/blob/fix-96/coffeeAGNTCY/coffee_agents/corto/deployment/helm/corto-ui/values.yaml#L20-L22, you can set dummy value or annotations if required.

PR #140, opened 2025-09-22
Fix: Prevent infinite loop and recursion limit in multi-tool-call scenarios

panyuenlau:
# Description

This PR introduces significant improvements to the agent's tool retry mechanism and error handling, making the system more robust and reliable, especially when interacting with external services or multiple A2A agents.

## Problem

Previously, the error handling is missing when building the agent graph. This caused the underlying error (for example, one of the agent apps goes offline) not getting detected. This caused the agent incorrectly decided to continue the conversation, leading to an infinitely loop until the recursion limit is hit, example error:

```
Operation failed: Recursion limit of 25 reached without hitting a stop condition. You can increase the limit by setting the `recursion_limit` config key.\nFor troubleshooting, visit: https://python.langchain.com/docs/troubleshooting/errors/GRAPH_RECURSION_LIMIT
```

And under the hood, the agents unintentionally send the same prompts to the LLM 25 times!

## Solution

This PR introduces error detection in agent nodes: the `_inventory_node` and `_orders_node` now perform comprehensive check for errors. When an error is detected, the system now directly reconstructs and re-sends the AIMessage containing the original tool calls that led to the failure. This retry is performed up to `MAX_TOOL_RETRIES` (default 3) times.

After all the retries, a high-level response will be returned to users:

Screenshot 2025-09-24 at 09 45 20

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
@@ -4,7 +4,7 @@
 import logging
 import os
 import uuid
-from typing import Optional
+from typing import Optional, List, Any, Dict
 from pydantic import BaseModel, Field
 
 from langchain_core.prompts import PromptTemplate
@@ -186,7 +186,9 @@ class ExchangeGraph:
                 logger.info(f"Reflection agent detected error message in last AIMessage: {last_message.content}. Ending conversation.")
                 return {
                     "next_node": END,
-                    "messages": [SystemMessage(content="Conversation ended due to tool error or service unavailability.")]
+                    "messages": [AIMessage(content="Conversation ended due to tool error or service unavailability. "
+                                                   "Please try again later or contact support if the problem persists.")
+                                 ]
                 }
 
         sys_msg_reflection = SystemMessage(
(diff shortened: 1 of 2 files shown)


coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
```diff
@@ -197,83 +229,265 @@ async def _inventory_node(self, state: GraphState) -> dict:
...
-            context = f"Tool responded: {tool_msg.content}"
+        # Find the AIMessage that initiated the tool calls
+        initiating_ai_message = None
+        for i in reversed(range(len(state["messages"]))):
+            msg = state["messages"][i]
+            if isinstance(msg, AIMessage) and hasattr(msg, 'tool_calls') and msg.tool_calls:
+                initiating_ai_message = msg
+                break
+
+        # Collect all ToolMessages that correspond to the tool calls from the initiating_ai_message
```
therealaditigupta: Out of genuine curiosity, why do we need to collect ToolMessages like this? Is it not possible to make the assumption that the last tool call in the last AIMessage would be the one with the error? Is this to protect against the case where the graph continues on after a message to another tool that "succeeds" therefore never catching the earlier tool mistake?

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
```diff
@@ -197,83 +229,265 @@ async def _inventory_node(self, state: GraphState) -> dict:
...
+                if msg is initiating_ai_message or isinstance(msg, HumanMessage):
+                    break
+            relevant_tool_responses.reverse()
+
+        is_error = False
+        aggregated_errors = []
+        error_keywords = ["error", "exception", "failed", "unavailable",
+                          "issue", "noresponderserror", "not recognized"]
+
+        if relevant_tool_responses:
```
therealaditigupta: https://langchain-ai.github.io/langgraph/how-tos/tool-calling/#handle-errors

Have we tried raising errors directly in the tool and seeing how LangGraph handles that? I see in the official documentation, raising vs returning will halt the graph at the error rather than recursing forever. In our current code (before these fixes) we simply return the error instead of raising it. That will allow us to not have to filter for errors manually.

therealaditigupta: Also I know you have looked at this for longer than I so apologies if I'm pointing out something you've already tried but wanted to get your take and experience

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
panyuenlau: The assumption that the last tool call in the `AIMessage` isn't correct. Example GraphState when one of the farms goes offline:

```json
{
  "inventory_tool_retry_count": 0,
  "last_inventory_tool_error": null,
  "last_orders_tool_error": null,
  "messages": [
    {
      "type": "HumanMessage",
      "content": "How much coffee does the Colombia and Vietnam farm have?",
      "additional_kwargs": {},
      "response_metadata": {},
      "id": "0c08ef8a-e839-4757-9e44-ee8c504f4bc8"
    },
    {
      "type": "AIMessage",
      "content": "",
      "additional_kwargs": {
        "tool_calls": [
          {
            "id": "call_1OMSdZNsyLDNHvGMtpoDX74r",
            "function": {
              "arguments": "{\"farm\": \"colombia\", \"prompt\": \"Please provide the current coffee yield inventory.\"}",
              "name": "get_farm_yield_inventory"
            },
            "type": "function"
          },
          {
            "id": "call_MTzqltB1eoT0A14qXEjhmY4Z",
            "function": {
              "arguments": "{\"farm\": \"vietnam\", \"prompt\": \"Please provide the current coffee yield inventory.\"}",
              "name": "get_farm_yield_inventory"
            },
            "type": "function"
          }
        ],
        "refusal": null
      },
      "response_metadata": {
        "token_usage": {
          "completion_tokens": 77,
          "prompt_tokens": 360,
          "total_tokens": 437,
          "completion_tokens_details": {
            "accepted_prediction_tokens": 0,
            "audio_tokens": 0,
            "reasoning_tokens": 0,
            "rejected_prediction_tokens": 0,
            "text_tokens": null
          },
          "prompt_tokens_details": {
            "audio_tokens": 0,
            "cached_tokens": 0,
            "image_tokens": null,
            "text_tokens": null
          }
        },
        "model_name": "gpt-4o-2024-08-06",
        "system_fingerprint": "fp_8458c98457",
        "id": "chatcmpl-CIgILZNgIHDCoO3dumTG9vYUCrBnU",
        "service_tier": null,
        "finish_reason": "tool_calls",
        "logprobs": null,
        "content_filter_results": {}
      },
      "id": "run--9c941040-68ab-4da5-a3b8-283cc246a3b7-0",
      "tool_calls": [
        {
          "name": "get_farm_yield_inventory",
          "args": {
            "farm": "colombia",
            "prompt": "Please provide the current coffee yield inventory."
          },
          "id": "call_1OMSdZNsyLDNHvGMtpoDX74r",
          "type": "tool_call"
        },
        {
          "name": "get_farm_yield_inventory",
          "args": {
            "farm": "vietnam",
            "prompt": "Please provide the current coffee yield inventory."
          },
          "id": "call_MTzqltB1eoT0A14qXEjhmY4Z",
          "type": "tool_call"
        }
      ],
      "usage_metadata": {
        "input_tokens": 360,
        "output_tokens": 77,
        "total_tokens": 437,
        "input_token_details": {
          "audio": 0,
          "cache_read": 0
        },
        "output_token_details": {
          "audio": 0,
          "reasoning": 0
        }
      }
    },
    {
      "type": "ToolMessage",
      "content": "Error: NoRespondersError()\n Please fix your mistakes.",
      "name": "get_farm_yield_inventory",
      "id": "3ebd3b55-effb-47cc-949c-7d39789b1fcf",
      "tool_call_id": "call_1OMSdZNsyLDNHvGMtpoDX74r",
      "status": "error"
    },
    {
      "type": "ToolMessage",
      "content": "45,000 pounds",
      "name": "get_farm_yield_inventory",
      "id": "9db70ca2-a0f7-49ad-a3d5-9383051ab847",
      "tool_call_id": "call_MTzqltB1eoT0A14qXEjhmY4Z",
      "status": "success"
    }
  ],
  "next_node": "inventory_broker",
  "orders_tool_retry_count": 0
}
```

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
panyuenlau: Actually, that's an excellent point and the best practice per LangGraph documentation! I will look more into updating the tool functions and simplifying the error handling logics in the agent.

Thanks for the pointer!

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
therealaditigupta: Ah I see! Thanks for the clarification and example GraphState

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
panyuenlau: updated implementation in https://github.com/agntcy/coffeeAgntcy/pull/140/commits/bb0c7033354285f1c76d7d65a5c1f0df0b8d942c

But unfortunately I couldn't find a way to make the retry log simpler than what I currently have...

Shridhar-2205 (dismissed):

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
```diff
@@ -197,83 +231,260 @@ async def _inventory_node(self, state: GraphState) -> dict:
...
+                    **state_update
+                }
+
+        prompt_parts = []
+        prompt_parts.append("""You are an orders broker for a global coffee exchange company. 
             Your task is to handle user requests related to placing and checking orders with coffee farms.
-            
+            """)
+
+        if is_error_detected and current_retry_count < self.MAX_TOOL_RETRIES and not last_ai_message_with_tool_calls:
```
shanchunyang0919: Hi @panyuenlau, thanks for your fix. While the error handling is definitely necessary, I feel the added system prompts make things overly complex for handling a simple retry issue. These prompts seem more suited for addressing business logic-related issues rather than retry logic.

The current implementation in the PR mixes business logic and error handling, and the nested structure could make debugging challenging—especially since many of these messages are generated by LLMs, where we can't always trust their accuracy.

I'm also unsure if asserting error strings against LLM responses is the best approach:
```
error_keywords = ["error", "exception", "failed", "unavailable", "issue", "noresponderserror", "offline", "unresponsive"]
```

Can we simplify this to make it more user-friendly and easier for external users to adopt? We should also consider that some users might want to copy the code directly into their existing codebase with minimal changes.

That said, I'll approve this PR if it's blocking your other tasks, but we should revisit this approach and the reflection node later.

shanchunyang0919 (dismissed):

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
panyuenlau: > for handling a simple retry issue.

Well, "simple retry issue" isn't really simple when it comes to agentic apps, see response from gemini:

Screenshot 2025-09-24 at 11 57 23

panyuenlau: > Can we simplify this to make it more user-friendly and easier for external users to adopt?

I would love to hear suggestions from the team if there is simpler solution to this infinite loop issue we are experiencing since I am not an expert with LangGraph and agentic apps.

But I do agree the solution I have in this PR is definitely more involved than I expected. Again, not sure if this due to my lack of experience

nandu-gh: Let's discuss during the standup.

shanchunyang0919 (approved):

Issue #186, opened 2025-10-13
[Bug]: Brazil Farm Orders Succeeding Despite Invalid Identity


therealaditigupta:
### Bug Description

The Brazil Farm agent is not registered in the Agent Identity Service. Within the Orders flow in the Supervisor graph, all order requests are validated to ensure that the target agent has a registered and valid badge within the Agent Identity Service. 

When an order is created in the Brazil farm, the request goes through despite the fact that the Brazil farm is not registered. Looking at the logs, I see

`exchange-server-lungo         | 2025-10-13 15:41:17,942 [WARNING] lungo.supervisor.tools: Identity verification failed for farm Brazil Coffee Farm: No matching app found, this farm probably does not have identity service enabled. Skipping identity verification.`

This should not be a warning but instead fail outright.

### Expected Behavior

When an order is created in the Brazil farm, the request should fail and an error should be returned mentioning the Invalid Badge or Registration.

### Affected Version

v1.0.0

### Steps to Reproduce

1. Spin up the Lungo agents
2. Run the prompt "I'd like to buy 200 lbs quantity of coffee at USD 500 price from Brazil"
3. Look at log output for the exchange-server-lungo

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

PR #189, opened 2025-10-13
BugFix: Improve farm identity verification by raising error when identity service is not enabled

therealaditigupta:
# Description

This fix addresses an issue in the farm identity verification process where the absence of a matching app (indicating the farm does not have identity service enabled) was previously logged as a warning and silently skipped. Now, the code raises an A2AAgentError with a clear error message and logs it as an error. This change ensures that identity verification failures are properly surfaced and handled, improving reliability and observability.

Fixes GitHub issue: [#186](https://github.com/agntcy/coffeeAgntcy/issues/186)

## Type of Change

- [X] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
@@ -129,10 +129,10 @@ def verify_farm_identity(identity_service: IdentityService, farm_name: str):
         matched_app = next((app for app in all_apps.apps if app.name.lower() == farm_name.lower()), None)
 
         if not matched_app:
-            logger.warning(f"Identity verification failed for farm {farm_name}: "
-                           f"No matching app found, this farm probably does not have identity service enabled. "
-                           f"Skipping identity verification.")
-            return
+            err_msg = f"Identity verification failed for farm {farm_name}: No matching app found, this farm does not have identity service enabled."
+            logger.error(err_msg)
+            raise A2AAgentError(err_msg)
+
 
         badge = identity_service.get_badge_for_app(matched_app.id)
         success = identity_service.verify_badges(badge)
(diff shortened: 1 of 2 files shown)


coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
```diff
@@ -129,10 +129,10 @@ def verify_farm_identity(identity_service: IdentityService, farm_name: str):
         matched_app = next((app for app in all_apps.apps if app.name.lower() == farm_name.lower()), None)
 
         if not matched_app:
-            logger.warning(f"Identity verification failed for farm {farm_name}: "
-                           f"No matching app found, this farm probably does not have identity service enabled. "
-                           f"Skipping identity verification.")
-            return
+            err_msg = f"Identity verification failed for farm {farm_name}: No matching app found, this farm does not have identity service enabled."
+            logger.error(err_msg)
+            raise A2AAgentError(err_msg)
```
panyuenlau, line 134: Would this cause any issues for the farms that don't have identity?

Asking since if we raise an error here, then the error will be caught in the reflection node, which is fine for the farms with identity, but for the ones that don't have identity, this will be still treated as an error and it will impact how the final response is constructed

therealaditigupta: We would want it to. We verify identity for all order requests to ensure only registered agents(farms) can be reached. If a farm doesn't have its identity registered, we want the flow to fail.

In this bug's case, there was no failure when an order was sent for an unregistered farm

panyuenlau: Yea, that makes sense for the order requests. 

But how about the inventory requests? If I remember correctly, we will hit the same error here. And this error will then be treated as a failure in the reflection node.

therealaditigupta: The inventory node does not call verify_farm_identity so it shouldn't be an issue

MrCsabaToth:
I tested this and I cannot buy from Brazil. Even though it returns a generic error, doesn't indicate that the reason is authentication, but probably that's OK

PR #383, opened 2026-02-24
Implement fail-fast in Lungo project in case the configured LLM doesn't support it

### LINKED ISSUE #247: Fail fast when LLM lacks streaming for agents that have native streaming support in their framework
**Description:**
Ensure agents with native streaming support fail fast if the configured LLM does not support streaming.

**Proposed Fix:**
Add a capability check during agent initialization; if streaming is required but unavailable, raise a clear configuration error and log guidance to switch models or use streaming LLM for that agent.

delthazor:
In Lungo project, agents treat native streaming support as a given, and if the configured LLM does not support that, it can cause problems. Therefore this capability should be detected early, cause a fail if missing.

# Description
Add an optional fail-fast check so that when the corresponding feature flag is enabled, the application does not start if the configured LLM does not support streaming (according to LiteLLM metadata), avoiding supervisors that require streaming from starting with a non-streaming model.

Feature flag: ENSURE_STREAMING_LLM (default false)

Check: read LLM_MODEL from config, call litellm.get_model_info() for supports_native_streaming; if the flag is on and the model does not support streaming, log guidance and raise StreamingNotSupportedError before the graph is created. Exception handling: known LiteLLM errors (e.g. NotFoundError, APIConnectionError) and any other exception are caught with debug logging; failures to get model info are treated as streaming not supported (fail closed).

Tests: unit tests (no app-sdk) and integration-style parametrized tests for all three supervisors.

## Issue Link
Fixes #247 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
@@ -65,21 +65,23 @@ class TestGetLlmStreamingCapability:
 
 
 class TestRequireStreamingCapability:
+  """Patch _config.ENSURE_STREAMING_LLM on the under-test module so require_streaming_capability sees it. patch() restores the original when the block exits so other tests are unaffected."""
+
   def test_early_return_when_flag_disabled(self):
     """When ENSURE_STREAMING_LLM is not true, require_streaming_capability returns without raising and does not call get_model_info."""
-    with patch(f"{_STREAMING_MODULE}.ENSURE_STREAMING_LLM", False):
+    with patch(f"{_STREAMING_MODULE}._config.ENSURE_STREAMING_LLM", False):
       with patch(f"{_STREAMING_MODULE}.litellm.get_model_info") as mock_get_model_info:
         require_streaming_capability("test_agent", _TEST_MODEL)
         mock_get_model_info.assert_not_called()
 
   def test_does_not_raise_when_capable(self):
-    with patch(f"{_STREAMING_MODULE}.ENSURE_STREAMING_LLM", True):
+    with patch(f"{_STREAMING_MODULE}._config.ENSURE_STREAMING_LLM", True):
       with patch(f"{_STREAMING_MODULE}.litellm.get_model_info", return_value={"supports_native_streaming": True}) as mock_get_model_info:
         require_streaming_capability("test_agent", _TEST_MODEL)
         mock_get_model_info.assert_called_once_with(model=_TEST_MODEL)
 
   def test_raises_streaming_not_supported_error_when_not_capable(self):
-    with patch(f"{_STREAMING_MODULE}.ENSURE_STREAMING_LLM", True):
+    with patch(f"{_STREAMING_MODULE}._config.ENSURE_STREAMING_LLM", True):
       with patch(f"{_STREAMING_MODULE}.litellm.get_model_info", return_value={"supports_native_streaming": False}) as mock_get_model_info:
         with pytest.raises(StreamingNotSupportedError) as exc_info:
           require_streaming_capability("test_agent", _TEST_MODEL)
@@ -90,7 +92,7 @@ class TestRequireStreamingCapability:
         mock_get_model_info.assert_called_once_with(model=_TEST_MODEL)
 
   def test_raises_when_get_model_info_raises(self):
-    with patch(f"{_STREAMING_MODULE}.ENSURE_STREAMING_LLM", True):
+    with patch(f"{_STREAMING_MODULE}._config.ENSURE_STREAMING_LLM", True):
       with patch(f"{_STREAMING_MODULE}.litellm.get_model_info", side_effect=Exception("unknown")) as mock_get_model_info:
         with pytest.raises(StreamingNotSupportedError) as exc_info:
           require_streaming_capability("other", _TEST_MODEL)
@@ -99,7 +101,7 @@ class TestRequireStreamingCapability:
         mock_get_model_info.assert_called_once_with(model=_TEST_MODEL)
 
   def test_optional_agent_name_stored_on_exception(self):
-    with patch(f"{_STREAMING_MODULE}.ENSURE_STREAMING_LLM", True):
+    with patch(f"{_STREAMING_MODULE}._config.ENSURE_STREAMING_LLM", True):
       with patch(f"{_STREAMING_MODULE}.litellm.get_model_info", return_value={"supports_native_streaming": False}) as mock_get_model_info:
         with pytest.raises(StreamingNotSupportedError) as exc_info:
           require_streaming_capability("", _TEST_MODEL)


codyhartsook (dismissed):
Implementation looks solid

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/agent.py
pregnor, line 1: **Question**: In case of recruiter pattern, shouldn't we also expect the remote agents to support it with the A2A agent evaluatior? (`RemoteAgentConnections` is using streaming configured client.)

coffeeAGNTCY/coffee_agents/lungo/common/streaming_capability.py
```diff
@@ -0,0 +1,51 @@
...
+    self.agent_name = agent_name
+    self.model = model
+    self.message = message
+    super().__init__(message)
+
+
+def get_configured_llm_streaming_capability() -> bool:
+  """Return True only if the configured LLM supports native streaming (LiteLLM static metadata)."""
+  try:
+    model_info = litellm.get_model_info(model=LLM_MODEL)
```
pregnor: **Question**: Wouldn't it be more versatile if we didn't use the config global here, but took it as an input parameter?
In that case the function is reusable for testing any model, not just the baked in default.

Same for the require function below.

coffeeAGNTCY/coffee_agents/lungo/common/streaming_capability.py
```diff
@@ -0,0 +1,51 @@
...
+  """Raised when the configured LLM does not support streaming but the agent requires it."""
+
+  def __init__(self, agent_name: str, model: str, message: str):
+    self.agent_name = agent_name
+    self.model = model
+    self.message = message
+    super().__init__(message)
+
+
+def get_configured_llm_streaming_capability() -> bool:
```
pregnor: **Question**: isn't it a better idea, to return `(bool, Exception)` here and allow `require_streaming_capability()` to "wrap" the exception there? That way even the stack trace can be preserved as the original.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
pregnor, line 1: **Optional**: Opinionated, but IMO a table driven test approach (like https://lorenzopeppoloni.com/tabledriventestspy/, or more so using [pytest's parametrization](https://docs.pytest.org/en/7.1.x/example/parametrize.html), see example uses on the previous project) would be effective here to eliminate some of the bloat and help readability between inputs and expectations.
The current solution is fine, and it could be just me feeling so, just a thought.

Edit: also similarly to what we do in `coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_fail_fast.py` just below.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
```diff
@@ -0,0 +1,95 @@
...
+      with patch("common.streaming_capability.litellm.get_model_info", side_effect=Exception("unknown")):
+        with pytest.raises(StreamingNotSupportedError):
+          require_streaming_capability("other")
+
+  def test_optional_agent_name_stored_on_exception(self):
+    with patch("common.streaming_capability.ENSURE_STREAMING_LLM", True):
+      with patch("common.streaming_capability.litellm.get_model_info", return_value={"supports_native_streaming": False}):
+        with pytest.raises(StreamingNotSupportedError) as exc_info:
+          require_streaming_capability("")
+        assert exc_info.value.agent_name == ""
```
pregnor, line 108: **Question**: Is there a reason the 3 exception cases are not using the same validation checks?
I would expect them to do so.
The reason for the comment was `test_raises_when_get_model_info_raises` not testing anything other than the correct exception class being instantiated, but then I noticed the inconsistency between the other 2 as well.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
```diff
@@ -0,0 +1,95 @@
...
+require_streaming_capability = _mod.require_streaming_capability
+
+from unittest.mock import patch
+
+import pytest
+
+
+class TestGetConfiguredLlmStreamingCapability:
+  def test_returns_true_when_supports_native_streaming_true(self):
+    with patch("common.streaming_capability.litellm.get_model_info", return_value={"supports_native_streaming": True}):
```
pregnor: **Question**: Shouldn't we also add `mock.assert_called_once()` in each cases to be sure the function callings are what we expect?

Same in the other test file.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/agent.py
delthazor: You are probably right, this feature could be used on all projects, where message streaming is in place, including future projects. Although, I'm not sure on how to port it to be a common utility, without reimplementing under each project. That is probably worth a discussion.

coffeeAGNTCY/coffee_agents/lungo/common/streaming_capability.py
delthazor: Sounds reasonable.
For me it seemed that there is only one model in use at any time, because everywhere it is referenced, the pre-configured one is returned, so I took a shortcut and checked only that one directly.

But of course the utility itself not have to be bound to this one value in particular, it could be parameterized to check any.

coffeeAGNTCY/coffee_agents/lungo/common/streaming_capability.py
delthazor: I'm not sure if that is needed for such a simple case, but that definitely would keep more information about the error. Might come in handy in case of network timeout or something like that causing the system to believe a model is not streaming capable when in fact it is.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
delthazor: According to OpenLLM, the possible API results for requesting metadata on a model is the following:
-If the LLM is recognized, and there is no error during the query, then we get a proper response (that can contain a false streaming compatibility flag)
-If the LLM is recognized but there is an error during query, the API returns the error
-If the LLM is not recognized (name misspelled, not existing, or referencing to proxy, etc.), the returned data may be empty, or incomplete.

I believe, Cursor generated the test cases for these three options, regardless of the fact, we make the same exception for each. The third one seems a bit arbitrary in that it defines an 'incomplete result' as only containing the on info we need anyways, and setting it to false so we would raise an exception anyways, so yeah, this probably needs some redesign... But the original intent is to test the scenario, when an incomplete metadata is returned due to incorrect model name in the query.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/common/test_streaming_capability.py
delthazor: Yes, that sounds correct.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/agent.py
pregnor: I had a misunderstanding on this, but we clarified it IRL, it's fine as it is.

codyhartsook (approved):

PR #395, opened 2026-03-02
Corto robustness fix for message timeout

### LINKED ISSUE #384: [Bug]: Integration tests for Corto seem flaky
### Bug Description

In rare cases the integration test package fails due to a timeout error in the following test:
test_sommelier.py::TestAuctionFlows::test_sommelier[ethiopia_profiles-SLIM]

The exchange agent asks “What are the flavor profiles of Ethiopian coffee?” over SLIM, but only receive a timeout response (SlimError.SessionError: message='receive timeout waiting for message'). The code then treats the failed response as if it had a body and does response.payload when response is None → Causes 'NoneType' object has no attribute 'payload' and error code 500.

Re-run of the test package usually solves the issue, and the test suite passes.

Here is an example of the erroneous behavior:
https://github.com/agntcy/coffeeAgntcy/actions/runs/22483051385/job/65125784638#step:9:1364

### Expected Behavior

The test suite should run consistently, with no corner- or edge-cases for occasional failures. Ideally timeout should not occur where it is not specifically required for the test case, and the code itself should be robust enough to handle accidental timeouts without failure.

### Affected Version

v1.0.0

### Steps to Reproduce

Since the error is occasional, there is no surefire way to reproduce the issue, but running Corto integration test suite multiple times, especially in slow environments can cause it to happen.

### Checklist

delthazor:
Add timeout error detections, retry and proper handling (along with corresponding tests) to Corto, because timeout errors caused test flakiness.

# Description

Corto integration tests fail due to occasional unexpected timeout errors which usually disappear after re-run. This solution aims to catch such cases, and handle them as good as possible, with minimal impact.
The first time a timeout is observed, the system tries a graceful retry. If the timeout happens again (or the second try fails with another type of error), it probably means a deeper problem that is out of scope for this change. In this case, the new solutions tries to relay as much info about the info towards the higher layers of the application, as possible, trying to avoid reading error messages like they were proper responses, and extracting none existent payloads.

## Issue Link

Fixes #384 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
+++ b/coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
@@ -1,6 +1,7 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
+import asyncio
 import logging
 from uuid import uuid4
 
@@ -85,38 +86,44 @@ def _parse_a2a_response(response):
     return None
 
 
+_A2A_MAX_ATTEMPTS = 5
+_A2A_BACKOFF_BASE = 3
+
+
 async def _send_a2a_with_retry(client, request):
-    try:
-        response = await client.send_message(request)
-        logger.info(f"Response received from A2A agent: {response}")
-        return response
-    except Exception as e:
-        if not _is_timeout_error(e):
-            if _is_no_payload_error(e):
-                raise RemoteAgentNoResponseError(
-                    "Remote agent returned no response (missing or invalid payload).",
-                    cause=e,
-                ) from e
-            logger.error("A2A send_message failed: %s", e)
-            raise
-        logger.warning("A2A request timed out, retrying once.")
+    """
+    Send request to A2A client. On timeout, retry up to 4 times (5 attempts total)
+    with exponential backoff (base 3, delays 1s, 3s, 9s, 27s). Only timeout errors
+    are retried; other errors propagate immediately.
+    """
+    for attempt in range(_A2A_MAX_ATTEMPTS):
         try:
             response = await client.send_message(request)
-            logger.info(f"Response received from A2A agent (retry): {response}")
+            logger.info("Response received from A2A agent.")
             return response
-        except Exception as e2:
-            if _is_timeout_error(e2):
-                raise TransportTimeoutError(
-                    "Remote agent did not respond in time (SLIM receive timeout).",
-                    cause=e2,
-                ) from e2
-            if _is_no_payload_error(e2):
-                raise RemoteAgentNoResponseError(
-                    "Remote agent returned no response (missing or invalid payload).",
-                    cause=e2,
-                ) from e2
-            logger.error("A2A send_message failed after retry: %s", e2)
-            raise
+        except Exception as e:
+            if not _is_timeout_error(e):
+                if _is_no_payload_error(e):
+                    raise RemoteAgentNoResponseError(
+                        "Remote agent returned no response (missing or invalid payload).",
+                        cause=e,
+                    ) from e
+                logger.error("A2A send_message failed: %s", e)
+                raise
+            if attempt < _A2A_MAX_ATTEMPTS - 1:
+                delay = _A2A_BACKOFF_BASE ** attempt
+                logger.warning(
+                    "A2A request timed out, retrying (attempt %s/%s) after %ss.",
+                    attempt + 2,
+                    _A2A_MAX_ATTEMPTS,
+                    delay,
+                )
+                await asyncio.sleep(delay)
+                continue
+            raise TransportTimeoutError(
+                "Remote agent did not respond in time (SLIM receive timeout).",
+                cause=e,
+            ) from e
 
 
 @agent(name="exchange_agent")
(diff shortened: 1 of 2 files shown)


coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
```diff
@@ -103,36 +109,39 @@ async def a2a_client_send_message(self, prompt: str):
...
         """
-        try:
-            factory = self.factory
-            a2a_topic = A2AProtocol.create_agent_topic(farm_agent_card)
-            transport = factory.create_transport(
-                DEFAULT_MESSAGE_TRANSPORT,
-                endpoint=TRANSPORT_SERVER_ENDPOINT,
-                # SLIM transport requires a routable name (org/namespace/agent) to build the PyName used for request-reply routing
-                name="default/default/exchange"
-            )
```
pregnor, line 116: **Question**: Why did we throw this comment away? We should inline it to the name field's line IMO.

coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
```diff
@@ -103,36 +109,39 @@ async def a2a_client_send_message(self, prompt: str):
...
+            params=MessageSendParams(
+                message=Message(
+                    message_id=str(uuid4()),
+                    role=Role.user,
+                    parts=[Part(TextPart(text=prompt))],
+                )
+            ),
+        )
+
+        def _parse_response(response):
```
pregnor: **Question**: Why are we inlining this function nested under the other function instead of promoting it to a top level function?

I would may be thinking about refactoring the control structure of the `a2a_client_send_message()` to be made up of multiple smaller functions instead of mixing levels of detail.

For example we could have separate helpers to 
 
- assemble the request, 
- parse the response
- implement the retry (especially for an exponential back-off that I advised offline)

And that way this higher level function would just be aggregating calls to those while conveying the higher level logic without the nitty gritty details crammed in place.

coffeeAGNTCY/coffee_agents/corto/exchange/errors.py
```diff
@@ -0,0 +1,60 @@
...
+    current = exc
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        if isinstance(current, SlimError.SessionError):
+            return True
+        current = getattr(current, "__cause__", None) or getattr(current, "__context__", None)
+    return False
+
+
+def _is_no_payload_error(exc: BaseException) -> bool:
```
pregnor, line 57: **Question**: Could you check for me if this makes the GPT5Pro error a bit more clear? I have seen exceptions on missing payloads with that one and I'm not sure if this has any impact.

It's ok to say no, I'm just curious.

coffeeAGNTCY/coffee_agents/corto/tests/unit/exchange/test_agent_timeout_retry.py
pregnor, line 1: **Question**: I might be wrong here, but couldn't this be made more concise with parametrization of the tests? (table testing)
Are the assert conditions so vastly different they can't be made generic?

coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
```diff
@@ -141,7 +150,37 @@ async def a2a_client_send_message(self, prompt: str):
...
+                        "Remote agent did not respond in time (SLIM receive timeout).",
+                        cause=e2,
+                    ) from e2
+                if _is_no_payload_error(e2):
+                    raise RemoteAgentNoResponseError(
+                        "Remote agent returned no response (missing or invalid payload).",
+                        cause=e2,
+                    ) from e2
+                logger.error("A2A send_message failed after retry: %s", e2)
+                raise
```
pregnor: **Question**: What do you think about using exponential backoff here, set for like 30s of timeout with a couple retries on a base of 3 for example (starting with the power of 0).

coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
delthazor: Put it back inline.

coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
delthazor: Extracted appropriate functions to distill the logic.

coffeeAGNTCY/coffee_agents/corto/exchange/errors.py
delthazor: I didn't see any improvements on that, sorry.

coffeeAGNTCY/coffee_agents/corto/tests/unit/exchange/test_agent_timeout_retry.py
delthazor: Changed the test to a more table-like parameterized approach.

coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
delthazor: Makes sense, so I added an example version of the exponential backoff retry.

PR #396, opened 2026-03-02
feat(ci): refactor GH workflow triggers and associated logic

### LINKED ISSUE #223: Optimize CI to Only Run Necessary Jobs Relevant to Commit
# Description

Currently, the CI triggers all jobs on each commit in PR, which is inefficient and results in waste of compute resources:

Image

# Proposed Solution

We should optimize the CI so it does not trigger the unnecessary actions, for example, making changes the code should not trigger the  "Helm Push Charts" action, making changes to charts shouldn't trigger the other 2 actions, etc.

### LINKED ISSUE #409: [Feature]: CI: make the GH workflow_dispatch trigger usable
### Problem Statement

Currently, in some GH workflows, the `workflow_dispatch` trigger is configured but is inconsistent with the conditionals in the rest of the workflow.

For example, we have in the [Docker workflow](https://github.com/agntcy/coffeeAgntcy/blob/0c7f244604b1dae8e3a76152a85be32d2634aba1/.github/workflows/docker-build-push.yaml):
```yaml
on:
  push:
    branches: [main]
    tags: ["*"]
  pull_request:
    branches: [main]
  workflow_dispatch:
```
but later on, when actually doing the Docker build:
```yaml
      - name: Build Docker image (PR)
        if: github.event_name == 'pull_request'
        uses: docker/build-push-action@v6
...
      - name: Build and push Docker image (main)
        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
        uses: docker/build-push-action@v6
...
      - name: Build and push Docker image (tag)
        if: startsWith(github.ref, 'refs/tags/')
        uses: docker/build-push-action@v6
...
```
which would lead to never doing anything on manual workflow runs (`workflow_dispatch` trigger).

The same applies to the [Helm workflow]( https://github.com/agntcy/coffeeAgntcy/blob/0c7f244604b1dae8e3a76152a85be32d2634aba1/.github/workflows/helm-push.yaml).

### Proposed Solution

Look it making proper use of the `workflow_dispatch` trigger.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

mihaialexandrescu:
# Description

This PR aims to optimize which CI jobs run, when and to what extent.

A few examples:
  * only Corto python code changes -> Lungo shouldn't docker build, helm push or run its integration tests; docs shouldn't rebuild.
  * only helm chart content changes for any subproject -> only the helm workflow should run and only for that subproject
  * only files relevant to docs change -> nothing other than docs build should happen
  * for PRs we shouldn't upload artifacts (this was already fine; just writing it to clarify the scenario)

I tried to not alter the actual build steps logic other than providing a different skeleton for the jobs so that we can have the differentiation between subprojects (in status quo they are all bundled up in a single matrix in multiple workflows). In the PR workflow code you will see a job per subproject.

I took the chance to update a few of the GH Actions versions.

I added a `tags: ["*"]` trigger for the helm workflow, similarly to what the docker workflow has.

When reviewing the `on:` triggers, keep in mind that, per GH workflow docs:
> Path filters are not evaluated for pushes of tags. 

The helm version I chose to update to is based on a "Note" at the top of the README of the project for `azure/setup-helm` currently pointing to that as a default.

Occasionally, I left in some debug statements because I believe that will make troubleshooting workflow runs just a bit easier.

I tested these changes on a fork of this repo.  
I couldn't run the integration tests since final test run relies on some secrets for LLM access but what's important there is that the conditionals work ok.

## Issue Link

Fixes #223 
Fixes #409 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/test-reusable.yaml
+++ b/.github/workflows/test-reusable.yaml
@@ -1,0 +1,75 @@
+name: Integration Tests per Project
+
+on:
+  workflow_call:
+    inputs:
+      project_dir:
+        description: 'The directory of the project to test.'
+        required: true
+        type: string
+      pip_overrides:
+        description: 'One-per-line PEP 508 specs to force (shared across demos; applied only if declared).'
+        required: false
+        type: string
+        default: ""
+      pip_constraints:
+        description: 'One-per-line constraints (shared across demos; applied only if declared).'
+        required: false
+        type: string
+        default: ""
+      docker_overrides:
+        description: 'One-per-line service=image[:tag] (shared across demos; applied only if declared).'
+        required: false
+        type: string
+        default: ""
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ${{ inputs.project_dir }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.13'
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Apply uv overrides/constraints (if applicable)
+        if: ${{ inputs.pip_overrides != '' || inputs.pip_constraints != '' }}
+        uses: ./.github/actions/apply-uv-overrides
+        with:
+          project_dir: ${{ inputs.project_dir }}
+          pip_overrides: ${{ inputs.pip_overrides }}
+          pip_constraints: ${{ inputs.pip_constraints }}
+
+      - name: Apply compose image overrides (if any)
+        if: ${{ inputs.docker_overrides != '' }}
+        uses: ./.github/actions/apply-container-overrides
+        with:
+          project_dir: ${{ inputs.project_dir }}/
+          compose_file: docker-compose.yaml
+          docker_overrides: ${{ inputs.docker_overrides }}
+
+      - name: Resolve & install
+        run: |
+          uv lock --upgrade
+          uv sync --extra dev
+      
+      - name: Create CI env file
+        run: |
+          cat > .env <<'EOF'
+          LLM_MODEL=${{ secrets.LLM_MODEL }}
+          AZURE_API_KEY=${{ secrets.AZURE_API_KEY }}
+          AZURE_API_BASE=${{ secrets.AZURE_API_BASE }}
+          AZURE_API_VERSION=${{ secrets.AZURE_API_VERSION }}
+          EOF
+
+      - name: Run integration tests
+        run: uv run pytest -s
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -31,6 +31,43 @@ on:
         required: false
         type: string
         default: ""
+      test_corto:
+        description: 'Whether to test the Corto project.'
+        required: false
+        type: boolean
+        default: true
+      test_lungo:
+        description: 'Whether to test the Lungo project.'
+        required: false
+        type: boolean
+        default: true
+  workflow_dispatch:
+    inputs:
+      pip_overrides:
+        description: 'One-per-line PEP 508 specs to force (shared across demos; applied only if declared).'
+        required: false
+        type: string
+        default: ""
+      pip_constraints:
+        description: 'One-per-line constraints (shared across demos; applied only if declared).'
+        required: false
+        type: string
+        default: ""
+      docker_overrides:
+        description: 'One-per-line service=image[:tag] (shared across demos; applied only if declared).'
+        required: false
+        type: string
+        default: ""
+      test_corto:
+        description: 'Whether to test the Corto project.'
+        required: false
+        type: boolean
+        default: true
+      test_lungo:
+        description: 'Whether to test the Lungo project.'
+        required: false
+        type: boolean
+        default: true
 
 env:
   CORTO_DIR: coffeeAGNTCY/coffee_agents/corto
@@ -50,16 +87,34 @@ jobs:
         uses: actions/checkout@v6
         with:
           fetch-depth: 0
-      
+
+      - name: Log event info
+        run: |
+          echo "event_name=${{ github.event_name }}"
+          echo "ref=${{ github.ref }}"
+          echo "ref_name=${{ github.ref_name }}"
+
+      - name: Log call or dispatch inputs
+        if: ${{ github.event_name == 'workflow_call' || github.event_name == 'workflow_dispatch' }}
+        run: |
+          echo "pip_overrides=${{ inputs.pip_overrides }}"
+          echo "pip_constraints=${{ inputs.pip_constraints }}"
+          echo "docker_overrides=${{ inputs.docker_overrides }}"
+          echo "test_corto=${{ inputs.test_corto }}"
+          echo "test_lungo=${{ inputs.test_lungo }}"
+
       - name: Detect changed paths for main sub-projects
         id: check-subprojects
         run: |
           test_corto=false
           test_lungo=false
 
-          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == refs/heads/main ]]; then
+          if [[ "${{ github.event_name }}" == "push" ]]; then
             test_corto=true
             test_lungo=true
+          elif [[ "${{ github.event_name }}" == "workflow_call" || "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            test_corto=${{ inputs.test_corto }}
+            test_lungo=${{ inputs.test_lungo }}
           elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
             base="${{ github.event.pull_request.base.sha }}"
             head="${{ github.event.pull_request.head.sha }}"
@@ -78,111 +133,22 @@ jobs:
           echo "test_corto=$test_corto" | tee -a $GITHUB_OUTPUT
           echo "test_lungo=$test_lungo" | tee -a $GITHUB_OUTPUT
 
-
   integration-tests-corto:
     needs: subprojects-to-test
     if: ${{ needs.subprojects-to-test.outputs.test_corto == 'true' }}
-    runs-on: ubuntu-latest
-
-    defaults:
-      run:
-        working-directory: ${{ env.CORTO_DIR }}
-        
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v6
-
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.13'
-
-      - name: Set up uv
-        uses: astral-sh/setup-uv@v7
-
-      - name: Apply uv overrides/constraints (if applicable) — ${{ env.CORTO_DIR }}
-        if: ${{ inputs.pip_overrides != '' || inputs.pip_constraints != '' }}
-        uses: ./.github/actions/apply-uv-overrides
-        with:
-          project_dir: ${{ env.CORTO_DIR }}
-          pip_overrides: ${{ inputs.pip_overrides }}
-          pip_constraints: ${{ inputs.pip_constraints }}
-
-      - name: Apply compose image overrides (if any) — ${{ env.CORTO_DIR }}
-        if: ${{ inputs.docker_overrides != '' }}
-        uses: ./.github/actions/apply-container-overrides
-        with:
-          project_dir: ${{ env.CORTO_DIR }}/
-          compose_file: docker-compose.yaml
-          docker_overrides: ${{ inputs.docker_overrides }}
-
-      - name: Resolve & install — ${{ env.CORTO_DIR }}
-        run: |
-          uv lock --upgrade
-          uv sync --extra dev
-      
-      - name: Create CI env file
-        run: |
-          cat > .env <<'EOF'
-          LLM_MODEL=${{ secrets.LLM_MODEL }}
-          AZURE_API_KEY=${{ secrets.AZURE_API_KEY }}
-          AZURE_API_BASE=${{ secrets.AZURE_API_BASE }}
-          AZURE_API_VERSION=${{ secrets.AZURE_API_VERSION }}
-          EOF
-
-      - name: Run integration tests — ${{ env.CORTO_DIR }}
-        run: uv run pytest -s
+    uses: ./.github/workflows/test-reusable.yaml
+    with:
+      project_dir: coffeeAGNTCY/coffee_agents/corto # cannot access env context here
+      pip_overrides: ${{ inputs.pip_overrides }}
+      pip_constraints: ${{ inputs.pip_constraints }}
+      docker_overrides: ${{ inputs.docker_overrides }}
 
   integration-tests-lungo:
     needs: subprojects-to-test
     if: ${{ needs.subprojects-to-test.outputs.test_lungo == 'true' }}
-    runs-on: ubuntu-latest
-
-    defaults:
-      run:
-        working-directory: ${{ env.LUNGO_DIR }}
-        
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v6
-
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.13'
-
-      - name: Set up uv
-        uses: astral-sh/setup-uv@v7
-
-      - name: Apply uv overrides/constraints (if applicable) — ${{ env.LUNGO_DIR }}
-        if: ${{ inputs.pip_overrides != '' || inputs.pip_constraints != '' }}
-        uses: ./.github/actions/apply-uv-overrides
-        with:
-          project_dir: ${{ env.LUNGO_DIR }}
-          pip_overrides: ${{ inputs.pip_overrides }}
-          pip_constraints: ${{ inputs.pip_constraints }}
-
-      - name: Apply compose image overrides (if any) — ${{ env.LUNGO_DIR }}
-        if: ${{ inputs.docker_overrides != '' }}
-        uses: ./.github/actions/apply-container-overrides
-        with:
-          project_dir: ${{ env.LUNGO_DIR }}/
-          compose_file: docker-compose.yaml
-          docker_overrides: ${{ inputs.docker_overrides }}
-
-      - name: Resolve & install — ${{ env.LUNGO_DIR }}
-        run: |
-          uv lock --upgrade
-          uv sync --extra dev
-      
-      - name: Create CI env file
-        run: |
-          cat > .env <<'EOF'
-          LLM_MODEL=${{ secrets.LLM_MODEL }}
-          AZURE_API_KEY=${{ secrets.AZURE_API_KEY }}
-          AZURE_API_BASE=${{ secrets.AZURE_API_BASE }}
-          AZURE_API_VERSION=${{ secrets.AZURE_API_VERSION }}
-          EOF
-
-      - name: Run integration tests — ${{ env.LUNGO_DIR }}
-        run: uv run pytest -s
+    uses: ./.github/workflows/test-reusable.yaml
+    with:
+      project_dir: coffeeAGNTCY/coffee_agents/lungo # cannot access env context here
+      pip_overrides: ${{ inputs.pip_overrides }}
+      pip_constraints: ${{ inputs.pip_constraints }}
+      docker_overrides: ${{ inputs.docker_overrides }}


.github/workflows/docker-build-push.yaml
```diff
@@ -2,53 +2,84 @@ name: Build and Push All Docker Images
...
+            build_recruiter=true
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            base="${{ github.event.pull_request.base.sha }}"
+            head="${{ github.event.pull_request.head.sha }}"
+            echo "Commits to compare: base=$base head=$head"
+
+            # git diff --quiet implies --exit-code (0 means no changes (no need to build), 1 means changes were found (need to build))
+            build_corto=$(git diff --quiet --ignore-blank-lines $base $head "$CORTO_DIR" 2>/dev/null && echo false || echo true)
+            build_lungo=$(git diff --quiet --ignore-blank-lines $base $head "$LUNGO_DIR" 2>/dev/null && echo false || echo true)
+            build_recruiter=$(git diff --quiet --ignore-blank-lines $base $head "$RECRUITER_DIR" 2>/dev/null && echo false || echo true)
```
pregnor: **Question**: :Wouldn't it have debugging value if instead of throwing away the output with `--quiet` we logged the diff here to be able to trace back what caused the binary output results?

.github/workflows/docker-build-push.yaml
```diff
@@ -2,53 +2,84 @@ name: Build and Push All Docker Images
...
 
+          # Debug resulting build flags
+          echo "Container images to build for subprojects: corto=$build_corto lungo=$build_lungo recruiter=$build_recruiter"
+
+          # Output build flags for subsequent use
+          echo "build_corto=$build_corto" >> $GITHUB_OUTPUT
+          echo "build_lungo=$build_lungo" >> $GITHUB_OUTPUT
+          echo "build_recruiter=$build_recruiter" >> $GITHUB_OUTPUT
+
+  docker-build-corto:
```
pregnor, line 166: **Question**: I'm guessing the answer, but I must ask to confirm and document here, was there no way to easily make the docker build generic for the different use-cases? (aside from a couple inputs that we check to confirm which one we are in)

We may have gone through this, forgive me my ignorance, but wouldn't the docker-build-cortom, docker-build-lungo and docker-build-recruiter be possible to be extracted into a reusable workflow of docker-build-push-subproject.yaml or something?

I feel like we are doing almost exactly the same in regards to lungo and recruiter while also very similar to Corto, most of the diffs could be extracted to input parameters (for the matrix setup for example). It's totally possible I don't see a blocker, I just feel like we have duplicated logic and thus risk for drift.

Same Q for the helm and integration test workflows. Maybe we could have a central dispatcher that evaluates whether a subproject needs to run - so we can deduplicate the subproject evaluation - then fire off subproject workflows for each subproject. We can base the parallelism on whether sequential test-docker-helm would be too long.

.github/workflows/docker-build-push.yaml
```diff
@@ -82,109 +113,344 @@ jobs:
...
+            image_tag=${{ steps.get_tag.outputs.tag_name }}
+            git_branch=${{ github.ref_name }}
+          elif [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/main" ]]; then
+            push_container=true
+            platforms=linux/amd64,linux/arm64
+            image_tag=latest
+            git_branch=${{ github.ref_name }}
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            push_container=false
+            platforms=linux/amd64
```
pregnor: **Question**: Is there a reason we don't build arm64 container on PRs? Isn't that a source of slipping arm64 related issues through the integration point?

Edit: I kind of see the reason to spare the resources and only do 1 platform, I leave it up to you, no hard leaning from me.

.github/workflows/docker-build-push.yaml
```diff
@@ -82,109 +113,344 @@ jobs:
...
+            image_tag=pr-${{ github.event.pull_request.number }}
+            git_branch=${{ github.head_ref }}
+          else
+            push_container=false
+            platforms=linux/amd64
+            image_tag=unknown
+            git_branch=${{ github.ref_name }}
+          fi
```
pregnor: **Question**: This means on workflow dispatch we won't build any container. Wouldn't it be better, to build the image with `{{ github.ref_name }}` but not push it? That way the workflow can be examined on manual starts while no artifact is created to bloat the registry.

(I was actually debating myself whether to push a commit image tag in this case, because that helps with quick fixes and testing, but seeing the latest hackerbot-claw developments I'm fine leaving it in non-pushing mode by default, we can always work around that should a real need arise even if time critical.)

.github/workflows/docker-build-push.yaml
```diff
@@ -2,53 +2,84 @@ name: Build and Push All Docker Images
...
+            build_recruiter=$(git diff --quiet --ignore-blank-lines $base $head "$RECRUITER_DIR" 2>/dev/null && echo false || echo true)
+          fi
 
+          # Debug resulting build flags
+          echo "Container images to build for subprojects: corto=$build_corto lungo=$build_lungo recruiter=$build_recruiter"
+
+          # Output build flags for subsequent use
+          echo "build_corto=$build_corto" >> $GITHUB_OUTPUT
+          echo "build_lungo=$build_lungo" >> $GITHUB_OUTPUT
+          echo "build_recruiter=$build_recruiter" >> $GITHUB_OUTPUT
```
pregnor: **Recommended**: We **SHOULD** consider doing the print and output saving in one step to not forget either.

This is a generic comment for all the occurrences.

```suggestion
          echo "build_corto=${build_corto}" | tee -a "$GITHUB_OUTPUT"
          echo "build_lungo=${build_lungo}" | tee -a "$GITHUB_OUTPUT"
          echo "build_recruiter=${build_recruiter}" | tee -a "$GITHUB_OUTPUT"
```

An article on recommending `tee` for GHA setting: https://huonw.github.io/blog/2025/02/ci-tee/ (ex Rust core team member author)

Also my personal test: (GITHUB_OUTPUT should work similarly, it's treated as a file/pipe/whatever for interstep transfer)

```bash
pregnor _ 68697 : PREGNOR-M-7MT6 @ - . /Users/pregnor [ microk8s / default ]
-> rm test.log && echo "---" && echo "build_corto=$TEST" | tee -a test.log && echo "---" && cat test.log

# Begin: 2026-03-04T08:59:40.853218000+01:00

---
build_corto=test
---
build_corto=test

# Executed command: rm test.log && echo "---" && echo "build_corto=$TEST" | tee -a test.log && echo "---" && cat test.log
# End: 2026-03-04T08:59:40.955427000+01:00
# Duration: 102ms209us
# Exit code: 0

pregnor _ 68697 : PREGNOR-M-7MT6 @ 1- . /Users/pregnor [ microk8s / default ]
-> echo "build_lungo=$TEST" | tee -a test.log && echo "---" && cat test.log

# Begin: 2026-03-04T09:00:12.229449000+01:00

build_lungo=test
---
build_corto=test
build_lungo=test

# Executed command: echo "build_lungo=$TEST" | tee -a test.log && echo "---" && cat test.log
# End: 2026-03-04T09:00:12.328901000+01:00
# Duration: 99ms452us
# Exit code: 0
```

.github/workflows/docker-build-push.yaml
```diff
@@ -82,109 +113,344 @@ jobs:
...
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build Docker image (PR)
-        if: github.event_name == 'pull_request'
+      - name: Build and conditional push Docker image
+        if: github.event_name == 'pull_request' || (github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/')))
```
pregnor: **Question**: Would this work for a workflow_dispatch event? I have some concerns we are duplicating the workflow trigger here which I feel like is redundant and error prone. If we go with the build-don't-push idea of mine for the workflow_dispatch mentioned above then I think this if should be `always()` and be omitted.

.github/workflows/docker-build-push.yaml
```diff
@@ -2,53 +2,84 @@ name: Build and Push All Docker Images
...
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            base="${{ github.event.pull_request.base.sha }}"
+            head="${{ github.event.pull_request.head.sha }}"
+            echo "Commits to compare: base=$base head=$head"
+
+            # git diff --quiet implies --exit-code (0 means no changes (no need to build), 1 means changes were found (need to build))
+            build_corto=$(git diff --quiet --ignore-blank-lines $base $head "$CORTO_DIR" 2>/dev/null && echo false || echo true)
+            build_lungo=$(git diff --quiet --ignore-blank-lines $base $head "$LUNGO_DIR" 2>/dev/null && echo false || echo true)
+            build_recruiter=$(git diff --quiet --ignore-blank-lines $base $head "$RECRUITER_DIR" 2>/dev/null && echo false || echo true)
+          fi
```
pregnor: **Question**: What happens for a workflow_dispatch event here? Shouldn't we be setting all the builds to true then and build-don't-push the images?

.github/workflows/docs.yaml
```diff
@@ -4,6 +4,14 @@ on:
...
     branches:
       - main
+    paths:
+      - 'README.md'
+      - 'CONTRIBUTING.md'
+      - 'CHANGELOG.md'
+      - 'CODE_OF_CONDUCT.md'
+      - 'LICENSE'
+      - 'SECURITY.md'
+      - 'assets/**'
```
pregnor, line 14: **Optional**: I **MAY** also add `docs/**` in a forward compatible manner if we start using it.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: There's a lot of interleaved lines in the diff so it's not clear but the short answer is that I tried to keep the build parameter decisions as they were and just do restructuring (there are small exceptions to this, though) so as not to pollute the scope of the PR/issue too much.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: I don't follow.  
Do you mean to add the `github.ref_name` as the `image_tag` for this fallthrough case?  
Otherwise, I don't think it's this bit of yaml that would make it so that we don't build any container on workflow dispatch, no?

.github/workflows/docker-build-push.yaml
pregnor: Ok, let's leave it as is.

.github/workflows/docker-build-push.yaml
pregnor: Sorry, my bad, I misread the code here, I think I took push_container for build_container instead, don't mind my comment, nothing to do here.

.github/workflows/docs.yaml
mihaialexandrescu: Done in 9b6123423446a78edcffdd6b1d72282b0f9bcb1e.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: Good idea!  
Done in 99b6d46aad98f563c99c19f8dc5e7478bc041eec.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: I agree.  
Done in 99b6d46aad98f563c99c19f8dc5e7478bc041eec with the refactor around the `dir_has_changes()` function.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: > Would this work for a workflow_dispatch event? I have some concerns we are duplicating the workflow trigger here which I feel like is redundant and error prone.

1. I don't know. But that is not a good thing and I think I/we should investigate this (maybe in a different issue?).
2. I wanted to focus on restructuring the workflow code instead of making new decisions on what the pre-existing logic is. What you see in the highlighted line is a compaction of the 3 distinct build steps we had (1 for each condition that is aggregated on this newly built line). I didn't really add or remove conditions - I tried to refine them to achieve the goal of the issue of not building everything all the time. I don't have enough reps with GH workflows to do something with deeper changes quickly so this is the compromise I chose.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: > What happens for a workflow_dispatch event here?

I don't know. We talked a bit offline. I aimed to restructure the pre-existing logic (to address the ticket more quickly), not change it.

> Shouldn't we be setting all the builds to true then and build-don't-push the images?

Within a PR, I don't think we want to build Lungo if only Corto changed or have a Corto change blocked because Lungo doesn't build. I don't think all build flags should be true all the time even if we wouldn't push.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: > I just feel like we have duplicated logic and thus risk for drift.

We talked about it offline.  
I agree. That is the downside of this version of structuring the workflows. The main bottleneck I'd say was my lack of reps for a long time with GH workflows (especially with reusable ones) and I wanted to get feedback more quickly on the rest of the logic inside the PR while also achieving what the triggering ticket was about.

I can take on that additional improvement (and yes, it extends to the other workflows too). It would be good practice.

mihaialexandrescu: In https://github.com/agntcy/coffeeAgntcy/blob/47ae2f2576973242862187b5955b5b717f12c23e/.github/workflows/docker-build-push.yaml#L504-L550 there is an idea on how to form the `matrix` dynamically (as JSON) and then import it later in a single build job. How do you feel about that idea?  (please note that the "code" around it has changed in the mean time in what you reviewed as part of this PR)

Alternative to that is to extract the build job into a different workflow file. Might be preferable.

I'm also guessing that the way to determine if a subproject has changed could be extracted into a separata GH Action that we'd maintain under the `.github/actions` folder ? One idea for that is in https://github.com/agntcy/coffeeAgntcy/blob/99b6d46aad98f563c99c19f8dc5e7478bc041eec/.github/workflows/docker-build-push.yaml#L52-L63

mihaialexandrescu: I pushed a working idea that employs a reusable workflow for:
 * helm in 943ee5f3d25d4f3200fad3704ec1c61f9c625538
 * docker in b2756461f1cd09d90f1412b6d9617e339a815fa8
 
Those commits also simplify the conditions for some triggers like push and build of subprojects by not repeating the conditions at the top of the GH workflow.  
For example:
```
[[ "${{ github.event_name }}" == "push" && ( "${{ github.ref }}" == refs/tags/* || "${{ github.ref }}" == refs/heads/main ) ]]
```
becomes
```
[[ "${{ github.event_name }}" == "push" ]]
```
and relies only on the `on` triggers to avoid accidental pushes to registries:
```
on:
  push:
    tags: ["*"]
    branches:
      - main
    ...
  pull_request:
    branches:
      - main
    ...
```

Let me know what you think.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: I tried always setting both platforms on the fork I tested other changes on and I don't see a significant difference in resource consumption (mainly time spent building containers). It is true I didn't run it with all the containers in Lungo.  
Commit b2756461f1cd09d90f1412b6d9617e339a815fa8 makes it so that we build for both platforms/multi-platform during PRs as well.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: The "per-project call site" for the reusable workflows like the `docker` one in b2756461f1cd09d90f1412b6d9617e339a815fa8 could be furthered reduced by using [YAML Anchors and Aliases](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases).  

It would look like this:
```
  docker-build-corto:
    needs: [subprojects-to-build, compute-build-parameters]
    if: ${{ needs.subprojects-to-build.outputs.build_corto == 'true' }}
    strategy:
      matrix:
        ... ...
    ... ...
    uses: ./.github/workflows/docker-build-reusable.yaml
    with: &docker-build-common-inputs     # <<< Defining the YAML Anchor
      name: ${{ matrix.image.name }}
      dockerfile: ${{ matrix.image.dockerfile }}
      push: ${{ needs.compute-build-parameters.outputs.push == 'true' }}
      platforms: linux/amd64,linux/arm64
      image_tag: ${{ needs.compute-build-parameters.outputs.image_tag }}
      git_branch: ${{ needs.compute-build-parameters.outputs.git_branch }}

  docker-build-lungo:
    needs: [subprojects-to-build, compute-build-parameters]
    if: ${{ needs.subprojects-to-build.outputs.build_lungo == 'true' }}
    strategy:
      matrix:
        ... ...
    ... ...
    uses: ./.github/workflows/docker-build-reusable.yaml
    with: *docker-build-common-inputs     # <<< calling on the anchor

  docker-build-recruiter:
    needs: [subprojects-to-build, compute-build-parameters]
    if: ${{ needs.subprojects-to-build.outputs.build_recruiter == 'true' }}
    strategy:
      matrix:
        ... ...
    ... ...
    uses: ./.github/workflows/docker-build-reusable.yaml
    with: *docker-build-common-inputs    # <<< calling on the anchor
```

I tested it now and it seems to work fine.

I don't know what to say about the readability. I am somewhat used to them but not everyone likes them.

pregnor: > there is an idea on how to form the matrix dynamically (as JSON) and then import it later in a single build job. How do you feel about that idea?

I'm fine with the JSON matrix being made beforehand, the only ask is to log it when it's assembled before it is executed.

> Alternative to that is to extract the build job into a different workflow file. Might be preferable.

That was my original expectation, but I'm not hardlined on it.

Whichever seems more intuitive from the uninitiated reader's perspective.

> I'm also guessing that the way to determine if a subproject has changed could be extracted into a separata GH Action that we'd maintain under the .github/actions folder ? One idea for that is in 

I haven't thought of this, but good idea. I thought we will have one initial workflow that determines which project changed and then calls out to each other workflow to do the heavy lifting for the changed projects only through workflow call inputs. That way we would define the project change evaluation only once and we wouldn't need the action.

> helm in https://github.com/agntcy/coffeeAgntcy/commit/943ee5f3d25d4f3200fad3704ec1c61f9c625538
>    docker in https://github.com/agntcy/coffeeAgntcy/commit/b2756461f1cd09d90f1412b6d9617e339a815fa8

Those look nice.

> YAML anchors

I like the end result, I just gotta get used to the full definition being at the first usage. Ask others' opinion as well please, but it's fine by me. This is the 2nd time I see YAML anchors used.

.github/workflows/docker-build-push.yaml
pregnor: Ok, let's revisit this in the final version.

.github/workflows/docker-build-push.yaml
pregnor: Ok, let's postpone fixing that, please create a ticket for it.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: I opened this https://github.com/agntcy/coffeeAgntcy/issues/409 issue to track the work on `workflow_dispatch` including flagging what I think is an issue currently. I could be wrong.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: I opened this https://github.com/agntcy/coffeeAgntcy/issues/409 issue to track the work on workflow_dispatch including flagging what I think is an issue currently. I could be wrong.

.github/workflows/docker-build-push.yaml
mihaialexandrescu: I'll get others' opinion on the Yaml Anchors topic.

In the mean time, I pushed updated versions of the reusable workflow idea:
  * helm in a67a133afd64151a2b94ed0540df5ab050d6b014
  * docker in 3b736e39f92f76f72514bd28e03625673c82c64e
  * integration tests (without recruiter, as we have it today) in 1b5159dda18695af07ecb33b4f53a866b465cd2b
  
Between them, these follow the same general pattern of workflow reuse.

delthazor: Regarding the anchors:
I see the added value, but agree with Patrik that it needs some adjusting for a reader that the full definition is at the first usage site. I could live with the above solution, but I think if YAML version supports merge keys, the structure might be improved like this:

```
x-docker-build-common-inputs: &docker-build-common-inputs
  push: ${{ needs.compute-build-parameters.outputs.push == 'true' }}
  platforms: linux/amd64,linux/arm64
  image_tag: ${{ needs.compute-build-parameters.outputs.image_tag }}
  git_branch: ${{ needs.compute-build-parameters.outputs.git_branch }}


jobs:
  docker-build-corto:
    needs: [subprojects-to-build, compute-build-parameters]
    if: ${{ needs.subprojects-to-build.outputs.build_corto == 'true' }}
    strategy:
      matrix: ...
    uses: ./.github/workflows/docker-build-reusable.yaml
    with:
      <<: *docker-build-common-inputs
      name: ${{ matrix.image.name }}
      dockerfile: ${{ matrix.image.dockerfile }}

```
Although I'm not entirely sure that really would work, but it worth a try.

mihaialexandrescu: That wouldn't work because this is not arbitrary YAML. We have to conform to Github Workflow schemas.  
The extent to which we can use the Anchors feature is described [here](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases).

Just in case, I tried it on the fork and the workflow errors out right away:

```yaml
common-inputs: &common-inputs     <<< Line 36
  push: ${{ github.event_name == 'push' || ( github.event_name == 'workflow_dispatch' && inputs.push ) }}
  build_corto: ${{ steps.check-subprojects.outputs.build_corto }}
  build_lungo: ${{ steps.check-subprojects.outputs.build_lungo }}
```
> Invalid workflow file: .........(Line: 36, Col: 1): Unexpected value 'common-inputs'
> 
> Invalid workflow file: .github/workflows/fake01-helm-push.yaml#L1
(Line: 36, Col: 1): Unexpected value 'common-inputs'

.github/workflows/docker-build-push.yaml
```diff
@@ -1,190 +1,261 @@
...
+      git_branch: ${{ steps.build_params.outputs.git_branch }}
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Log event info
+        run: |
+          echo "event_name=${{ github.event_name }}"
+          echo "ref=${{ github.ref }}"
+          echo "ref_name=${{ github.ref_name }}"
```
pregnor, line 127: **Question**: Is it intentional we are logging the constant event info in every job instead of giving it their dedicated job and removing that from the rest of the workflows?

I am all for it if we feel like we want to put dynamic data in here that is job dependent, but this seems a bit redundant now, because under a single workflow you will see the same values anyway, but if we feel like it's easier to debug if we don't have to check the dedicated job for these info then I can let it fly.

Especially in relation to the next step which is also a workflow level setting, not dynamic between jobs.

Same Q for the Helm push workflow.

.github/workflows/test-reusable.yaml
```diff
@@ -0,0 +1,75 @@
...
+        uses: ./.github/actions/apply-container-overrides
+        with:
+          project_dir: ${{ inputs.project_dir }}/
+          compose_file: docker-compose.yaml
+          docker_overrides: ${{ inputs.docker_overrides }}
+
+      - name: Resolve & install
+        run: |
+          uv lock --upgrade
+          uv sync --extra dev
```
pregnor, line 63: **Question**: Wait a second, this uv lock --upgrade is modifying the versioned dependencies potentially creating `uv.lock` file changes (upgrades allowed by the constraints) and causing dirty state which differs from the committed state.
And also the following `uv sync --extra dev ` causes the dependencies to be updated in the environment as well meaning the tests will run with this new dependency tree.

```bash
# uv lock --help
  -U, --upgrade                                        Allow package upgrades, ignoring pinned versions in any existing output file. Implies `--refresh`
      --refresh                            Refresh all cached data
```

> `uv lock --upgrade` updates **only the lockfile**, not your installed environment.
> 
> More specifically:
> 
> - It re-solves your dependencies and **updates `uv.lock` to the newest versions allowed by the constraints in `pyproject.toml`** (for all packages, unless you restrict it with extra flags).  
> - It **does not install or upgrade** anything in your virtualenv/system; it just refreshes the “snapshot” of exact versions recorded in the lockfile.  
> - To actually install those new versions into your environment, you run `uv sync` afterwards, which makes the environment match the updated `uv.lock`.
> 
> So the usual workflow is:
> 
> ```bash
> uv lock --upgrade   # refresh uv.lock with latest allowed versions
> uv sync             # bring the environment in line with uv.lock
> ```
> 
> Sources
> [1] Understanding the Difference Between uv lock --upgrade and uv sync in Python Dependency Management https://www.youtube.com/watch?v=I3ExkyBY4YA
> [2] What's the difference between uv lock --upgrade and uv sync? https://stackoverflow.com/questions/79475051/whats-the-difference-between-uv-lock-upgrade-and-uv-sync
> [3] uv: An In-Depth Guide to Python's Fast and Ambitious New Package ... https://www.saaspegasus.com/guides/uv-deep-dive/
> [4] `uv lock` empties lockfile when upgrading python · Issue #13951 · astral-sh/uv https://github.com/astral-sh/uv/issues/13951
> [5] Does `uv lock —upgrade` not do what you want? - Hacker News https://news.ycombinator.com/item?id=44358537
> [6] Locking and syncing | uv - Astral Docs https://docs.astral.sh/uv/concepts/projects/sync/
> [7] What is the intended workflow for updating dependencies with uv? https://github.com/astral-sh/uv/issues/6692
> [8] Working on projects | uv - Astral Docs https://docs.astral.sh/uv/guides/projects/
> [9] UV — Intuitively and Exhaustively Explained https://iaee.substack.com/p/uv-intuitively-and-exhaustively-explained
> [10] How to use a uv lockfile for reproducible Python environments https://pydevtools.com/handbook/how-to/how-to-use-a-uv-lockfile-for-reproducible-python-environments/

I don't think this is desirable, because ideally

1. We wouldn't want CI to modify the committed code even in CI-run-time to have reproducible builds and also to have tight control on the supply chain (exact locked versions used).
2. We would want to use the same dependency versions in the local development environment that is used by the CI for testing the changes.

The only reason I'm not posting a changes required comment here is because I'm hoping I'm missing something that invalidates my logic.

I do see this is a legacy "feature" in the CI, but if you agree I would create a ticket and have this tackled IMMEDIATELY after this PR, because this is a serious issue IMO.

pregnor (approved):
LGTM, good job on the refactor

.github/workflows/docker-build-push.yaml
mihaialexandrescu: That step is there intentionally in all top-level refactored workflows. I also intentionally avoided adding it in the reusable workflows because I wanted to stay away from using trigger-related information in those.

My reason for it is to help troubleshooting. We have some decision making happening around that place that centers around which type of trigger we had and I favored showing the information within the same workflow page (as opposed to in another workflow which means another page in the GH Workflows UI) and close to the spot using it.

If we end up not liking it when using it, I won't resist removing these from the workflows and centralizing that log step.

.github/workflows/test-reusable.yaml
mihaialexandrescu: You are absolutely right! That's no good.  
Thank you for catching this!

> I do see this is a legacy "feature" in the CI, but if you agree I would create a ticket and have this tackled IMMEDIATELY after this PR, because this is a serious issue IMO.

Agreed. If you create the issue, you can assign it to me right away and I'll deal with it.

.github/workflows/docker-build-push.yaml
pregnor: I just don't get it why we need the same 3 static infos in every job where they are determined on a workflow trigger level.
There could be a trigger log job with this one step for example.

But it's not that big of a deal, so I can let this go, I just shared my way of thinking, I don't push to remove these, just asked why repeat redundant information.

.github/workflows/test-reusable.yaml
pregnor: https://github.com/agntcy/coffeeAgntcy/issues/415

PR #413, opened 2026-03-07
fix(Security): security hardening (bugs and preventive measures)

### LINKED ISSUE #400: [Security]: Security issues & improvements
### Problem Statement

1. target="_blank" / window.open without noopener (tab-napping)
* Where: ExternalLinkButton uses window.open(url, "_blank") with no noopener.
* Risk: The new tab can access window.opener and redirect the original tab (tab-napping).

2. URL used in ExternalLinkButton / window.open
* Where: ExternalLinkButton is generic; currently only used with grafanaSessionUrl (app-controlled). If any caller ever passes a user- or API-controlled URL, that URL could be javascript: or data: and lead to XSS when opened.

3. ReactMarkdown and XSS
* Where: InfoButton renders infoContent with <ReactMarkdown>{infoContent}</ReactMarkdown>.
* Risk: If infoContent ever comes from the API or user input, malicious markdown/HTML could lead to XSS (depending on ReactMarkdown and plugin config).

4. Discovery / graph node data and githubLink
* Where: data.githubLink is used in CustomNode and transportNode as href. Today it’s set from config (graphConfigs) or from updateTransportLabels (API returns transport type; URL is built from config). Discovery-built nodes use templates from config for githubLink.
* Risk: If in the future we set githubLink (or any link) from discovery/API payload without validation, a javascript: or data: URL could be used.

5. CSRF
* Where: Frontend sends cookies via credentials: "include" / withCredentials on POST/fetch.
* Risk: CSRF is mainly a backend concern; the frontend does not implement CSRF tokens.

6. Sensitive data in logs
* Where: logger.apiError logs error.message and error.stack; several places log full responses or API/stream data (e.g. ChatArea response, IdentityApi/DirectoryApi nodeData, MainArea discovery event, streaming stores logging raw lines on parse failure).
* Risk: In production, logs could expose PII, session/agent data, or internal stack traces.

7. Error messages shown to the user
* Where: User-facing error text comes from error.response?.data?.detail (and similar) via parseApiError / parseFetchError.
* Risk: If the backend ever returns HTML or script inside detail/message, and that string is ever rendered as HTML, it could cause XSS.

8. JSON.parse of localStorage
* Where: App.tsx does JSON.parse(localStorage.getItem(...)) for saved messages.
* Risk: If an attacker could write to localStorage (e.g. via XSS elsewhere), they could inject malicious JSON; the parsed data is then rendered as message content. We already render message content as text, so risk is limited to abuse of app logic, not direct DOM XSS.

### Proposed Solution

1. target="_blank" / window.open without noopener (tab-napping)
* Fix: Prefer a normal `<a href={url} target="_blank" rel="noopener noreferrer">` instead of window.open
(If for some reason we keep window.open, use the third parameter so the new window is created with noopener behavior (e.g. window.open(url, "_blank", "noopener,noreferrer")). Ensure every target="_blank" and window.open in the app uses noopener/noreferrer.

2. URL used in ExternalLinkButton / window.open
* Recommendation: For any URL that can be user- or API-controlled, validate scheme (e.g. allow only https: and optionally http: for known hosts) before passing to ExternalLinkButton or window.open. Keep grafanaSessionUrl and other app-built URLs as the only sources for that button unless you add this validation.

3. ReactMarkdown and XSS
* Recommendation: Ensure infoContent is only from trusted app/constants. If we ever feed API or user content into it, use a sanitization layer (e.g. rehype-sanitize with a strict schema) or render that content as plain text instead.

4. Discovery / graph node data and githubLink
* Recommendation: If any link URL can come from API or user data, validate scheme (e.g. allow only https:) and optionally allowlist host/domain before using it in href or window.open.

5. CSRF
* Recommendation: Check backend to protect state-changing requests: SameSite cookie attributes (e.g. SameSite=Strict or Lax), and/or CSRF tokens / double-submit cookie. Ensure our backend does one of these; no front-end change required for CSRF beyond following backend’s token usage if it adds one.

6. Sensitive data in logs
* Recommendation: In production, avoid logging full API responses, raw NDJSON lines, and node/agent payloads; log only endpoint, status, and minimal identifiers. (Let's check if this is feasible.) Consider not logging full stack traces to client-accessible logs, or redacting them. Remove or guard console.log/console.debug of request/response/data with import.meta.env.DEV or similar.

7. Error messages shown to the user
* Recommendation: Always render these messages as text (e.g. in React, avoid dangerouslySetInnerHTML for them). Ensure backend does not put HTML or script in detail/message. Optionally sanitize or strip tags before display if the backend is not fully trusted.

8. JSON.parse of localStorage
* Recommendation: Keep message content rendered as text only. Optionally wrap JSON.parse in try/catch and treat invalid or unexpected shape as “no messages” to avoid crashes or unexpected UI from malformed data.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

misi-bp:
# Description

This PR applies **security hardening** to the Lungo frontend: XSS mitigations, safe logging, and safer handling of untrusted or persisted data.

**Changes:**

- **XSS / client-side security**
  - **External links:** Use an `<a>` with `rel="noopener noreferrer"` instead of a button for external URLs so new windows cannot access the opener.
  - **Safe external URL usage:** Ensure external links (e.g. directory, Grafana) are opened safely.
  - **GitHub links:** Make GitHub links XSS-safe (no `javascript:` or unsafe `href`).
  - **Error messages:** User-facing error text from `parseApiError` / `parseFetchError` (and in ChatArea) is taken from a single path and run through a **stripHtml** helper (DOMPurify in the browser) so backend HTML in `detail`/`message` is never rendered as HTML. **InfoButton** still runs `stripHtml` before passing content to ReactMarkdown; rehype-sanitize remains for markdown output.
  - **Pure-text display:** `stripHtml` was removed from display paths that only use React text children (`{string}`), relying on React's default escaping there.
  - **ESLint:** Add `no-restricted-syntax` to forbid `dangerouslySetInnerHTML` and `innerHTML` so future code cannot accidentally render unsanitized HTML.

- **Logging**
  - **Production:** The main `logger` is a no-op in production (no logs to the client).
  - **Diagnostics:** `unsafeLogger` is available for rare prod diagnostics; it runs payloads through a redaction layer and is flagged by ESLint. Use only for non-sensitive data.

- **Persisted data**
  - **localStorage messages:** Reading saved chat messages from localStorage is wrapped in try/catch and validated: only valid JSON that is an array of `Message`-shaped objects is used; otherwise initial messages are `[]`. Prevents crashes and misuse from malformed or tampered data.

- **Developer experience**
  - **Lungo frontend:** Add a single `npm run check` script that runs lint check, format check, and typecheck (with a short comment in `package.json`).

**Scope:** Lungo frontend only (no Corto changes in this branch).

## Issue Link

Fixes #400 

## Type of Change

- [X] Bugfix
- [X] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useAppChatState.ts
+++ b/coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useAppChatState.ts
@@ -13,12 +13,21 @@ export interface UseAppChatStateParams {
   selectedPattern: PatternType
 }
 
+/** Safe read of persisted messages; invalid JSON or non-array is treated as no messages. */
+function getInitialMessages(): Message[] {
+  try {
+    const saved = localStorage.getItem(LOCAL_STORAGE_KEY)
+    if (!saved) return []
+    const parsed: unknown = JSON.parse(saved)
+    return Array.isArray(parsed) ? (parsed as Message[]) : []
+  } catch {
+    return []
+  }
+}
+
 /** Chat UI state, messages persistence, and response/input handlers. */
 export function useAppChatState({ selectedPattern }: UseAppChatStateParams) {
-  const [messages, setMessages] = useState<Message[]>(() => {
-    const saved = localStorage.getItem(LOCAL_STORAGE_KEY)
-    return saved ? JSON.parse(saved) : []
-  })
+  const [messages, setMessages] = useState<Message[]>(getInitialMessages)
 
   const [aiReplied, setAiReplied] = useState<boolean>(false)
   const [buttonClicked, setButtonClicked] = useState<boolean>(false)


coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/const.ts
```diff
@@ -100,16 +101,50 @@ export type ApiErrorInfo = {
...
+    "&#39;": "'",
+    "&apos;": "'",
+  }
+  for (const [ent, char] of Object.entries(entities)) {
+    out = out.split(ent).join(char)
+  }
+  out = out.replace(/&#(\d+);/g, (_, n) => String.fromCharCode(Number(n)))
+  out = out.replace(/&#x([0-9a-fA-F]+);/g, (_, n) =>
+    String.fromCharCode(parseInt(n, 16)),
+  )
```
delthazor, line 135: Just being curious here:
You replay have to do this by hand, if no proper DOM is available? No library function or 3rd party sanitizer is available for cases like this? I mean this looks like pretty standard stuff, so I'm surprised that you have to do it manually yourself.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/logger.ts
```diff
@@ -1,20 +1,98 @@
...
+  "set-cookie",
+  "x-api-key",
+  "x-auth-token",
+  "credential",
+  "credentials",
+  "privatekey",
+  "private_key",
+  "ssn",
+  "social_security",
+])
```
delthazor, line 44: Same question again: This seems like a pretty standard list. Of course there is value in having control over the list, but this typically looks like something that should have been already done by someone whose job is to make extensive sanity checks on sensitive data.

But once again: I'm just trying to learn from this, so no need to change your solution.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/const.ts
misi-bp: I'm sorry, I don't quite understand the question: 
In the frontend the most spread 3PP for this task is DOMPurify. The stripHTML function above by default uses DOMPurify, and the lines you marked here are only the fallback. DOMPurify can do its job using a browser's DOM parser (without it, the sanitized snippet cannot be reinserted into an html tag). The reason behind is DOMPurify is not a string-based sanitizer, because those can break easily. All in all: what you were missing (a 3PP) is already there, and this is just a nice and flexible fallback. Not a must-have one, I admit it, but this is a shared function so why not.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/logger.ts
misi-bp: No worries, I like even debates and really appreciate any review or feedback.

So this is a different case, then the one above. Frankly, I had to lookup if any package contains a standard list of sensitive keys to see how they can handle this. The only one I've found is 'sensitive-fields', which is a list-only package. However, I don't see the point to replace our list with that one --> for debugging reasons this list should remain editable, flexible, so one extra package for this case is unnecessary IMO.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/const.ts
delthazor: Maybe I was't clear enough, I see this fallback as a good option, I'm just surprised that there is no string-based sanitizer that can already blacklist these special chars, and regex patterns, so you don't have to do this on your own. Anyways, the DOMPurify will stand its ground in any real environment, so I think the solution is good as-is.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/logger.ts
delthazor: I agree, if there is no added benefit in a 3rd party package, that would beat having full control of the list (or at least an option to easily add or remove elements from that 3rd party list) then there is no value in just adding another dependency.
Still I'm surprised that despite the really much ado about cybersecurity, there are no official (and by that I mean professionally supervised) solution for the most simplest cases.

Issue #387, opened 2026-02-27
[Bug]: docker compose up can fail on startup sequence


pregnor:
### Bug Description

Lungo `docker compose up --build --detach --force-recreate` reliably repros `Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint lungo-mce-api-layer (7a0877b80f26a9ad8d349a41e2fb36aba82028fe7f4eca0ff7f6b0f556da79b8): failed to bind host port 0.0.0.0:8080/tcp: address already in use`.

Same for Corto `docker compose up --build --detach --force-recreate`: `Error response from daemon: failed to set up container networking: driver failed programming external connectivity on endpoint corto-mce-api-layer-1 (fe2af9db20b3370cef952399bf87c80a0f038f8ca05e1d020793cce508f726c8): failed to bind host port 0.0.0.0:8080/tcp: address already in use`

Without `--force-recreate`, this usually only happens on cold starts (`--build`).

```bash
pregnor _ 3158 : - @ - . /Users/pregnor/development/src/github.com/agntcy/coffeeAgntcy/coffeeAGNTCY/coffee_agents/lungo (main %|u=) [ microk8s / default ]
-> docker compose up --build

# Begin: 2026-02-25T14:53:44.254197000+01:00

WARN[0000] The "OPENAI_ENDPOINT" variable is not set. Defaulting to a blank string.
WARN[0000] The "OPENAI_MODEL_NAME" variable is not set. Defaulting to a blank string.
WARN[0000] The "OPENAI_API_KEY" variable is not set. Defaulting to a blank string.
[+] Building 16.7s (16/16) FINISHED
 => [internal] load local bake definitions                       0.0s
 => => reading from stdin 633B                                   0.0s
 => [internal] load build definition from Dockerfile.ui          0.0s
 => => transferring dockerfile: 935B                             0.0s
 => [internal] load metadata for docker.io/library/node:20-alpi  0.7s
 => [internal] load .dockerignore                                0.0s
 => => transferring context: 2B                                  0.0s
 => [1/9] FROM docker.io/library/node:20-alpine@sha256:09e2b3d9  0.0s
 => [internal] load build context                                0.1s
 => => transferring context: 2.15MB                              0.1s
 => CACHED [2/9] WORKDIR /app                                    0.0s
 => CACHED [3/9] RUN apk add --no-cache wget curl iputils        0.0s
[...]

PR #414, opened 2026-03-09
Improve Lungo startup time

### LINKED ISSUE #386: [Bug]: Slow Lungo startup times without indication
### Bug Description

Currently Lungo is starting up very slowly.
Full cold start is around 13m, logistics hot start is around 3m30s.
During these wait times the docker compose containers are already running, the UI is already showing the graphs but the suggested prompts are still not available, while there is no indication of this happening or how long it will take.

### Expected Behavior

There should be either blocking dependency or user feedback about ongoing initialization before Lungo UI elements can be used.

It should be investigated whether this can be optimized both in startup time, sequence and user feedback.

### Affected Version

v1.0.0

### Steps to Reproduce

1. Set up Lungo according to the tutorial with the correct profiles.
2. Fire off the docker compose and as soon as the UI is ready try to open the suggested prompts dropdown.
3. Observe the logs to see how long it takes to initialize the recruiter-supervisor, logistics-supervisor or auction supervisor agent before the suggested prompts are available on the UI.

### Checklist

delthazor:
# Description

Lungo project had 3 major issues that made it slow to start:

1:
Uvicorn auto reload was always on, causing significantly slower startup times. This is absolutely not needed unless developing backend features locally with immediate feedbacks. 

Solution:
New config flag introduced in .env to control Uvicorn hot reload. It defaults to false, but .env.example has it set to true, so if someone makes a fresh config by copying it, it will act as originally did. By setting it to false, boot time can get significantly better.

2:
Dependency definitions in the docker compose file were not aligned with real dependencies, thus boot order sometimes got messed up, causing some components to start with connection errors, timeouts and retries, slowing their boot up, os straight up failing to boot.

Solution:
Dependencies were aligned with real boot requirements, along with health-checks where applicable, making boot more robust, and somewhat quicker by not waiting for retries.

3:
Supervisors often waited for the full agent graph to be inited, before responding to anything, but some lite-queries (like health checks, and config queries read from static files, e.g.: suggested prompts) could be served regardless.

Solution:
Init of the agent graph was moved into a background process, making the agent be able to respond to queries that can be completed without it. This is not making any change on total boot time, but the system can get responsive much earlier, giving users and smoother experience, that feels quicker.

## Issue Link

Fixes #386 
Possibly have a positive impact on #387 (needs confirmation)

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/tests/integration/docker_helpers.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/tests/integration/docker_helpers.py
@@ -74,8 +74,20 @@ def up(files: List[str], services: List[str]):
 
 def down(files: List[str]):
     # 'down' ignores service list; it tears down the whole project
-    cmd = _compose_cmd(files) + ["down", "-v"]
+    cmd = _compose_cmd(files) + ["down", "-v", "--remove-orphans"]
     _run(cmd)
+
+
+def remove_container_if_exists(container_name: str) -> None:
+    """Force-remove a container by name if it exists. Use after down() to clear leftovers from other runs."""
+    res = subprocess.run(
+        ["docker", "rm", "-f", container_name],
+        capture_output=True,
+        text=True,
+        cwd=PROJECT_DIR,
+    )
+    if res.returncode == 0:
+        print(f"> Removed leftover container {container_name}")
 
 def _container_id(files: List[str], service: str) -> str:
     cmd = _compose_cmd(files) + ["ps", "-a", "-q", service]
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/main.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/main.py
@@ -190,6 +190,16 @@ async def handle_stream_prompt(request: PromptRequest, req: Request):
         raise HTTPException(status_code=400, detail=str(ve))
     except Exception as e:
         raise HTTPException(status_code=500, detail=f"Operation failed: {str(e)}")
+
+
+@app.get("/ready")
+async def ready(req: Request):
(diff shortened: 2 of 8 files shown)


coffeeAGNTCY/coffee_agents/lungo/tests/integration/docker_helpers.py
```diff
@@ -74,9 +74,21 @@ def up(files: List[str], services: List[str]):
...
+def remove_container_if_exists(container_name: str) -> None:
+    """Force-remove a container by name if it exists. Use after down() to clear leftovers from other runs."""
+    res = subprocess.run(
+        ["docker", "rm", "-f", container_name],
+        capture_output=True,
+        text=True,
+        cwd=PROJECT_DIR,
+    )
+    if res.returncode == 0:
+        print(f"> Removed leftover container {container_name}")
```
pregnor, line 90: **Note**: I feel like in the long run it would be better to build a library and a task automation out of these infrastructure manipulation. I have a really bad feeling of everything touching the docker context from everywhere (dev, test, etc.).

delthazor: Actually, I don't think all these tests should really use docker at all. The application should be independent of the platform used, and even if it worth testing wether it can run in a docker environment or not, that should be just a single test scenario, that only checks "docker up - app healthcheck - docker down". All other tests should be running independently. I see how that is not feasible right now, but in the long run this could be a reasonable goal.

pregnor: +1

PR #417, opened 2026-03-10
test: fix warnings & turn warnings into errors

pregnor:
# Description

Treat pytest warnings as errors and fix or work around deprecations so the test suite runs clean with `filterwarnings = ["error"]` in **corto**, **lungo**, and **recruiter**. Prefer fixing root causes; add narrow, documented filters only where the warning comes from third-party code we can’t change.

### Corto
- **Starlette HTTP_413:** In `tests/conftest.py`, set `starlette.status.HTTP_413_REQUEST_ENTITY_TOO_LARGE = starlette.status.HTTP_413_CONTENT_TOO_LARGE` before any import that pulls in a2a, so a2a’s import never hits the deprecated `__getattr__` (no warning, no ignore).
- **LiteLLM:** Bump to `>=1.82.0` for Pydantic ConfigDict in litellm types.
- **pytest-asyncio:** Bump to `>=0.23.0,<0.24` so event-loop handling uses the modern asyncio API and avoids “There is no current event loop” / “Event loop is closed” under Python 3.13.
- Pytest options moved from `pytest.ini` into `pyproject.toml`; `filterwarnings = ["error"]`.

### Lungo
- **Starlette HTTP_413:** Same conftest patch as corto in `tests/conftest.py`.
- **pytest-asyncio:** Bump to `>=0.23.0,<0.24` for Python 3.13–compatible event-loop behavior.
- Pytest options moved from `pytest.ini` into `pyproject.toml`; `filterwarnings = ["error"]`.

### Recruiter
- **Pydantic:** Use `ConfigDict(use_enum_values=False)` in `AgentEvalConfig` instead of nested `class Config` (Pydantic v2).
- **Logging:** Avoid `llm_response.model_dump()` in evaluator callbacks; use a small, logging-safe repr to prevent Message/Choices serializer deprecation warnings.
- **Runner/plugins:** Use `App(name=..., root_agent=..., plugins=...)` and `Runner(app=app, session_service=...)` instead of deprecated `Runner(plugins=...)`.
- **LiteLLM:** Bump to `>=1.82.0` for ConfigDict fix in litellm types.
- **Third-party (google-adk):** Two targeted `filterwarnings` entries: `run_config.save_input_blobs_as_artifacts` (runners) and `MCPTool` (mcp_tool) until ADK is updated; comments in `pyproject.toml` explain and say to remove when ADK is fixed.
- Pytest config moved into `pyproject.toml`; `filterwarnings = ["error"]` plus the two ADK ignores above.

## Issue Link

- (This was a prerequisite for being able to properly verify other changes in other tickets.)

## Type of Change

- [X] Other (please describe): developer experience, fix test warnings.

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/tests/conftest.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/tests/conftest.py
@@ -1,0 +1,9 @@
+# Apply before any test or module import that pulls in a2a (e.g. agents.supervisors.recruiter).
+# a2a-sdk imports starlette.status.HTTP_413_REQUEST_ENTITY_TOO_LARGE, which starlette
+# exposes via __getattr__ with a DeprecationWarning. Defining the attribute here so it
+# exists on the module means a2a's import never triggers __getattr__, so no warning.
+import starlette.status  # noqa: E402
+
+starlette.status.HTTP_413_REQUEST_ENTITY_TOO_LARGE = (  # noqa: E402
+    starlette.status.HTTP_413_CONTENT_TOO_LARGE
+)


coffeeAGNTCY/coffee_agents/recruiter/pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
     "loguru>=0.7.3",
     "pydantic>=2.11.4",
     "python-dotenv>=1.1.0",
-    "litellm>=1.80.5",
+    "litellm>=1.82.0",  # 1.82+ includes Pydantic ConfigDict fix (was class Config in anthropic.py)
```
delthazor, line 13: This version change was added to all the other projects as well, without the comment. I don't think this version upgrade (or any other for that matter) needs commenting. I think only downgrades should be explained.

pregnor: I will do this in a separate PR.

PR #438, opened 2026-03-12
Lungo - Separation of telemetry so it can be switched off

### LINKED ISSUE #131: [Feature]: Allow option to disable/enable observability stack
### Problem Statement

While running apps locally or via docker compose, we need an option to enable/disable observability stack which has additional components like otel, grafana, clickhouse. Apps should work without such components.

### Proposed Solution

Add disable/enable option for observability stack which is confirmed with ioa-observer-sdk.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

delthazor:
# Description

Telemetry related docker containers collected to separate "observability" profile, which needs to be added to "COMPOSE_PROFILES" if someone needs them.

Also, OTEL provides, an official OTEL_SDK_DISABLED environment variable, which is respected by most libs that use OTEL, so it is used in our code, to determine if tracing is needed or not. 

If solution proves useful/handful it can be extended to Corto.

## Issue Link

Fixes #131 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/tests/integration/conftest.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/tests/integration/conftest.py
@@ -5,7 +5,12 @@
 Pytest fixtures using the simple ProcessRunner.
 Replace prior xprocess usage with these.
 """
+import atexit
 import os
+
+# Force OTel SDK on in tests so we get recording spans; without this NonRecordingSpan would be created (from disabled SDK) which has no .attributes.
+os.environ["OTEL_SDK_DISABLED"] = "false"
+
 import re
 import time
 import sys
@@ -18,7 +23,6 @@ from tests.integration.docker_helpers import up, down, remove_container_if_exist
 from tests.integration.process_helper import ProcessRunner 
 
 LUNGO_DIR = Path(__file__).resolve().parents[2]
-print("LUNGO_DIR:", LUNGO_DIR)
 
 AGENTS = {
     # auction agents
@@ -62,6 +66,8 @@ _ACTIVE_RUNNERS = []
 # ---------------- utils ----------------
 
 def _base_env():
+    # Use test env: SDK on so spans are recording (avoids NonRecordingSpan.attributes error).
+    otel_disabled = os.environ.get("OTEL_SDK_DISABLED", "false")
     return {
         **os.environ,
         "PYTHONPATH": str(LUNGO_DIR),
@@ -70,6 +76,7 @@ def _base_env():
         "OTEL_SDK_DISABLED": "false",
         "PYTHONUNBUFFERED": "1",
         "PYTHONFAULTHANDLER": "1",
+        "OTEL_SDK_DISABLED": str(otel_disabled),
     }
 
 def _purge_modules(prefixes):
@@ -105,9 +112,25 @@ def _wait_ready(client, path, timeout_s=30.0, poll_s=0.5):
     raise RuntimeError(f"Ready check {path} did not return 200 within {timeout_s}s")
 
 # ---------------- session infra ----------------
+# docker_helpers passes env=os.environ to compose so infra containers use the same env as the test (e.g. OTEL_SDK_DISABLED).
 files = ["docker-compose.yaml"]
 if Path("docker-compose.override.yaml").exists():
     files.append("docker-compose.override.yaml")
+
+_session_docker_torn_down = False
+
+
+def _teardown_session_docker():
+    """Run docker compose down so session containers (slim, nats, grafana, etc.) are stopped. Idempotent."""
+    global _session_docker_torn_down
+    if _session_docker_torn_down:
+        return
+    _session_docker_torn_down = True
+    print("--- Tearing down session Docker (slim, nats, otel-collector, clickhouse, grafana) ---")
+    down(files)
+
+
+atexit.register(_teardown_session_docker)
 
 
 def _shutdown_otel_sdk():
@@ -144,8 +167,10 @@ def orchestrate_session_services():
     setup_identity()
     print("--- Session level service setup complete. Tests can now run ---")
     yield
-    _shutdown_otel_sdk()
-    down(files)
+    try:
+        _shutdown_otel_sdk()
+    finally:
+        _teardown_session_docker()
 
 def setup_transports():
     _startup_slim()
--- a/coffeeAGNTCY/coffee_agents/lungo/tests/integration/docker_helpers.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/tests/integration/docker_helpers.py
@@ -26,13 +26,22 @@ def _compose_cmd(files: List[str]) -> List[str]:
             cmd += ["-f", str(compose_file)]
     return cmd
 
+def _compose_env():
+    """Env for docker compose subprocess: use current process env so test-set vars (e.g. OTEL_SDK_DISABLED) are used."""
+    return dict(os.environ)
+
+
 def _run(cmd: List[str]):
     """
     Execute a docker compose command from PROJECT_DIR to make calls location-agnostic.
+    Passes current process env so compose (and variable substitution) uses the same env as the test.
     """
+    compose_env = _compose_env()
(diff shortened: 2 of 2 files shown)


coffeeAGNTCY/coffee_agents/lungo/config/config.py
```diff
@@ -6,6 +6,12 @@
 
 load_dotenv()  # Automatically loads from `.env` or `.env.local`
 
+# OTEL_SDK_DISABLED: common var used by 3rd-party libs built on OpenTelemetry (OTEL). When true, tracing is disabled.
+_otel_sdk_disabled_raw = os.getenv("OTEL_SDK_DISABLED", "false").strip().lower()
+OTEL_SDK_DISABLED = _otel_sdk_disabled_raw in ("true", "1", "yes")
+if OTEL_SDK_DISABLED:
+    os.environ.pop("OTLP_HTTP_ENDPOINT", None)
+
```
pregnor, line 14: **Question**: There is a common practice to avoid negative boolean names, because when negated it becomes double negative which can be convoluted to understand, for example `not DISABLE_DISALLOWLIST`.

Is there a reason we couldn't use `OTEL_SDK_ENABLED`?
Or rather `USE_OTEL_SDK` for example.
Or with a heavy heart following the `ENABLE_HTTP` convention already existing, `ENABLE_OTEL_SDK`.

Is it because we want the undefined env var default empty value to be "enabling" the OTEL SDK?
In this case, I would simply set the default in code to true and explicitly document it so an explicit false setting is needed to disable it.

Edit: OH FFS, now I'm reading the comment, I can't believe this... Ok whatever, then I guess this is the world we live in.

Is it like devilish to define our own name, read both ours and the common one, reconcile those two so if ours is defined then ours take precedence, but if only the common is defined we take the common and set the resulting value to the common one downstream? (and in case neither is defined, we do the common default which is disabled false which is enabled) Or am I overthinking this?

coffeeAGNTCY/coffee_agents/lungo/docs/group_conversation.md
```diff
@@ -55,7 +55,7 @@ User Prompt
 
 To enable A2A Group communication over SLIM, you need to run the SLIM message bus gateway.
 
-Additionally, run the observability stack that includes OTEL Collector, Grafana, and ClickHouse DB. You can do this by executing the following command:
+With profile-based Compose, set `COMPOSE_PROFILES` to include `observability` and `OTEL_SDK_DISABLED=false` in `.env`. Otherwise, start the observability stack explicitly:
```
pregnor: **Required**: This sentence makes no sense to me, we **MUST** fix that.

If we set `COMPOSE_PROFILES` to `observability` and `OTEL_SDK_DISABLED=false` that means we want to start the stack and enable the telemetry (disabled=false, I told you above this is bad practice). So there is no point in the "Otherwise" part, that's not otherwise, that's a consequence of those settings observability stack will start then.

IMO we should either say "set the compose profile and either implicitly or explicitly set the disabling to false to start the stack, otherwise disable the compose profile and set the env var disabled=true to disable telemetry completely" or say "The compose profile and a non-disabled telemetry sending is needed to start the telemetry. If you want to disable it, both remove it from the compose profile and set the env var disabled=true".

I leave the phrasing up to you as long as it describes the proper states.

coffeeAGNTCY/coffee_agents/lungo/docs/group_conversation.md
```diff
@@ -162,4 +162,4 @@ Expected output:
 
 ## Observability
 
-See [Observability](../README.md#observability)
+Observability requires both the `observability` profile in `COMPOSE_PROFILES` and `OTEL_SDK_DISABLED=false`. See [Observability](../README.md#observability).
```
pregnor: **Note**: See my previous comment, this is a false claim, `OTEL_SDK_DISABLED=false` is optional, because the code sets the env var reading default to false which will mean telemetry sending is enabled by default on the sending part, not the stack part, the compose profile is obviously needed.

TUTORIAL.md
```diff
@@ -367,6 +369,8 @@ share updates dynamically.
 
 ### 4. Inspect Traces in Grafana
 
+
+Observability works only when both the `observability` profile (in `COMPOSE_PROFILES`) and `OTEL_SDK_DISABLED=false` are set.
```
pregnor: **Note**: This is also related to my comment about the `OTEL_SDK_DISABLED=false` is not a hard requirement because of the existing default logic in the env var value reading.

TUTORIAL.md
```diff
@@ -298,11 +298,13 @@ All workshop services are containerized — start everything with one command:
 docker compose up --build
 ```
 
+The observability stack (Grafana, OTEL Collector, ClickHouse) is started when the `observability` profile is included in `COMPOSE_PROFILES` (e.g. in `.env`); set `OTEL_SDK_DISABLED=false` when you want telemetry. Using the profile without the env var means no telemetry is sent; using the env var without the profile can cause log noise and failed exports.
```
pregnor: **Note**: This is kind of okay here, but I would still slightly reword this, because of the soft requirement on the `OTEL_SDK_DISABLED=false` case and rather mention the explicit requirement on disabling the telemetry.

coffeeAGNTCY/coffee_agents/lungo/README.md
```diff
@@ -134,6 +134,8 @@ Before you begin, ensure the following tools are installed:
 
 Update your .env file with the provider model, credentials, and OTEL endpoint.
 
+**Observability:** To use the observability stack (Grafana, OTEL Collector, ClickHouse), set `COMPOSE_PROFILES` to include `observability` (e.g. `farms,logistics,recruiter,observability`) and set `OTEL_SDK_DISABLED=false` (or leave unset). To run without the stack, omit `observability` from `COMPOSE_PROFILES` and set `OTEL_SDK_DISABLED=true`. Using the observability profile without enabling telemetry (`OTEL_SDK_DISABLED=true`) means the stack runs but no telemetry is sent; using telemetry (`OTEL_SDK_DISABLED=false`) without the observability profile can cause log noise and failed telemetry exports.
```
pregnor: **Note**: See, here you used the correct phrasing for the default case.

coffeeAGNTCY/coffee_agents/lungo/config/config.py
pregnor, line 14: **Note**: Right now we can't do much to disable telemetry reporting on sender side when observability stack is not set up properly - maybe we could introduce another env var and check here, but that's somewhat cumbersome - but when we refactor the config interface this should be made way easier to use with a single aggregate setting regarding the stack and the sending because it's too much detail on the first time user to set this up properly.

pregnor (changes requested):

coffeeAGNTCY/coffee_agents/lungo/config/config.py
delthazor: I was thinking about introducing a simple "TRACING_ENABLED" env var, but I ran into some problems on how to auto trigger the Otel one, since the 3rd party stuffs we use are depending on that one.
As it turned cumbersome to keep the two sources of truth somehow in think, I thought we could more easily live by with this improperly named 3rd party var instead of always trying to do the bookkeeping between ours and theirs.

coffeeAGNTCY/coffee_agents/lungo/docs/group_conversation.md
delthazor: The otherwise parts refers to the command that follows (originally part of the document) which lists the corresponding containers one-by-one instead of calling them by profile name. Regardless I'll find a better wording.

coffeeAGNTCY/coffee_agents/lungo/config/config.py
pregnor: Yeah makes sense, thanks for trying anyway.

PR #441, opened 2026-03-12
feat: move paths_have_changes logic to dedicated GH Action

mihaialexandrescu:
# Description

This PR aims to move into its own GH Action a bit of Bash that we are using to determine if there are changes inside a subproject (or more recently a set of paths, including folders).  
This Bash logic is very similar currently across 3 GH workflows and may be reused for other future GH workflows that want to use that condition for PRs to limit the number of workflows that get run.

The GH Action starts with a validation step that checks that the input paths are valid JSON and that it is an array of strings. Nothing else (including nesting) is allowed for that input to pass the validation.

The workflows that use this action follow the action folder as a trigger too.

## Issue Link

Didn't create one. This is a follow up of an idea from this thread https://github.com/agntcy/coffeeAgntcy/pull/396#discussion_r2889192787. The particular Bash changed a lot but the idea remained.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/actions/check-paths-have-changes/action.yaml
+++ b/.github/actions/check-paths-have-changes/action.yaml
@@ -1,0 +1,59 @@
+name: Check filesystem paths for changes
+description: Check whether a list of filesystem paths have changes between two commits
+inputs:
+  base_sha:
+    description: The base SHA to compare against
+    required: true
+  head_sha:
+    description: The head SHA to compare against
+    required: true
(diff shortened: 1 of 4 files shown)


delthazor (approved):
lgtm

PR #444, opened 2026-03-16
(feat) Bring over timeout handling to Lungo

delthazor:
# Description

Corto already got a timeout handling functionality for network messages, that can be used for Lungo as well.
This solution was expanded by retry for "no-answer" messages, which actually cover internal SLIM timeouts in case of recipient taking long to answer.

## Issue Link
-

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/a2a_retry.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/a2a_retry.py
@@ -70,42 +70,62 @@ logger = logging.getLogger(__name__)
 
 async def send_a2a_with_retry(client, request):
     """
-    Send request to A2A client. On timeout, retry up to 4 times (5 attempts total)
-    with exponential backoff (base 3, delays 1s, 3s, 9s, 27s). Only timeout errors
-    are retried; other errors propagate immediately.
+    Send request to A2A client. On timeout or no response (None / no_payload), retry
+    up to 4 times (5 attempts total) with exponential backoff (base 3, delays 1s, 3s,
+    9s, 27s). Valid response with empty content is returned as-is.
     """
     for attempt in range(_A2A_MAX_ATTEMPTS):
         try:
             response = await client.send_message(request)
-            if response is None or getattr(response, "root", None) is None:
-                raise RemoteAgentNoResponseError(
-                    "Remote agent returned no response (missing or invalid payload).",
-                    cause=None,
-                )
-            return response
-        except RemoteAgentNoResponseError:
-            raise
-        except Exception as e:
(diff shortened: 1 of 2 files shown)


coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/a2a_retry.py
```diff
@@ -0,0 +1,131 @@
...
+    return False
+
+
+def _is_no_payload_error(exc: BaseException) -> bool:
+    """True iff the exception is an AttributeError for missing 'payload' (e.g. access on None)."""
+    return isinstance(exc, AttributeError) and getattr(exc, "name", None) == "payload"
+
+
+_A2A_MAX_ATTEMPTS = 5
+_A2A_BACKOFF_BASE = 3
```
codyhartsook, line 66: should we consider making these configurable? Prod cluster may want different settings than local dev?

delthazor: If the need arise, I think we can extract it later, but for now, I would keep it this way.

These numbers were chosen by calculating the last backoff time to be less then 30 seconds (27 in this case) so the total time we spend retrying is less then a minute.
Setting numbers higher would cause really huge waiting times, when the app actually appears to be hanging. 
Setting them much lower can run into the issue that the cause of the timeout is not resolved yet, so the workflow can fail, despite that it could have a clear path a few seconds later.

If another environment differs so greatly that different numbers are needed, then we can easily extract these few constants into env or similar config space, but for a start I would keep them as is.

coffeeAGNTCY/coffee_agents/lungo/tests/integration/test_auction.py
```diff
@@ -88,6 +91,30 @@ def load_auction_prompt_cases():
...
+    ):
+        resp = auction_supervisor_client.post(
+            "/agent/prompt",
+            json={"prompt": "What is the inventory of coffee in Brazil?"},
+        )
+    assert resp.status_code == 200
+    data = resp.json()
+    assert "response" in data
+    response_text = data["response"].lower()
+    assert "issue" in response_text or "try again" in response_text or "fail" in response_text or "communicat" in response_text
```
pregnor, line 115: **Note**: This condition seems a bit broad to me on the expectation side.

delthazor: Opened separate PR for the test update:
https://github.com/agntcy/coffeeAgntcy/pull/471

PR #472, opened 2026-03-18
fix(corto,fe): fix agent records

### LINKED ISSUE #391: [Bug]: Corto UI AGNTCY buttons lead nowhere
### Bug Description

Current Corto UI AGNTCY buttons are leading to a non-existent agent directory service instance.

We should fix those links.

Image

Image

Image

### Expected Behavior

The AGNTCY links in Corto should work.

### Affected Version

v1.0.0

### Steps to Reproduce

1. Set up Corto
2. Fire up Corto
3. On the Corto UI, click one of the AGNTCY buttons either for the Buyer or the Sommelier.

### Checklist

pregnor:
# Description

Fixed the Corto agent record button, it was pointing to a non-existing directory instance, now agent records are inlined locally like they are for lungo auction and logistics.

- Added generated OASF records for the exchange and farm agents.
- Added an `/agents/{agent-slug}/oasf` endpoint to the exchange API server.
- Added an integration test for the new endpoint to be returning the proper records.
- Updated the frontend to use the new backend endpoint for the agent record showing and display the content in a modal.
- Added playwright dependency and integration tests for testing the frontend functionality of the agent record buttons. Gitignored the playwright test output files.

## Issue Link

Fixes #391 

## Type of Change

- [X] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/corto/exchange/frontend/package.json
+++ b/coffeeAGNTCY/coffee_agents/corto/exchange/frontend/package.json
@@ -11,8 +11,8 @@
     "preview": "vite preview",
     "format": "prettier --write \"src/**/*.{js,jsx,ts,tsx,json,css,md}\"",
     "format:check": "prettier --check \"src/**/*.{js,jsx,ts,tsx,json,css,md}\"",
-    "test:e2e": "playwright test",
-    "test:e2e:ui": "playwright test --ui"
+    "test:integration": "playwright test",
+    "test:integration:ui": "playwright test --ui"
   },
   "dependencies": {
     "@radix-ui/react-select": "^2.2.5",


coffeeAGNTCY/coffee_agents/corto/exchange/frontend/src/components/MainArea/Graph/CustomNode.tsx
```diff
@@ -37,12 +43,28 @@ const CustomNode: React.FC<CustomNodeProps> = ({ data }) => {
     dark: agentDirectoryIconDark,
   })
 
+  const handleAgentDirectoryClick = (e: React.MouseEvent) => {
```
misi-bp: To prevent infinite loops in the future, it's worth to wrap each function defined in components in useCallback or useMemo. (If a function does not have side effects, feel free to move them outside the component, that's the safest way to prevent it, but useCallback is also fine in these cases as well. Please double check the other functions as well.)

misi-bp: Otherwise lgtm :)

pregnor: Fixed in [44cb627](https://github.com/agntcy/coffeeAgntcy/pull/472/commits/44cb6274faa63e0baf8b39f62f4439981b613f1e).

I only touched the frontend parts which were modified in this PR due to the release, we can fix this later when we are focusing on Corto.

I will also resolve the conversation because of the release and you being on PTO today, if there is anything left to do, let me know when you are back and we will do it.

coffeeAGNTCY/coffee_agents/corto/exchange/main.py
```diff
@@ -113,6 +112,29 @@ async def get_prompts():
     logger.exception(f"Failed to load suggested prompts: {str(e)}")
     raise HTTPException(status_code=500, detail=f"Failed to load prompts: {str(e)}") from e
 
+
+@app.get("/agents/{slug}/oasf")
+async def get_agent_oasf(slug: str):
+  """
+  Returns the OASF JSON for the specified agent slug from the static files.
+  """
+  oasf_path = Path(__file__).resolve().parent.parent / "oasf" / "agents" / f"{slug}.json"
```
delthazor: I'm not an expert on the matter, but is this slug safe to use? I mean, can it point to a file that exists, but not the intended ones?
I can't really find where it is originates from, and wether it can be overridden by typing a manual url, but it can be my oversight.

coffeeAGNTCY/coffee_agents/corto/exchange/frontend/src/components/MainArea/Graph/CustomNode.tsx
```diff
@@ -37,12 +51,31 @@ const CustomNode: React.FC<CustomNodeProps> = ({ data }) => {
...
+        ).getBoundingClientRect()
+        const position = {
+          x: buttonRect.left + buttonRect.width / 2,
+          y: buttonRect.bottom + 12,
+        }
+        data.onOpenOasfModal(data, position)
+      }
+    },
+    [data],
+  )
```
delthazor, line 70: The data object may change due to re-renders assigning a new reference (even if members don't change), causing the useCallback having reduced efficiency.
If you only depend on the member values it worth to list those as dependencies, or if you have a trustable id (like data.slug) that is guaranteed to change if and only when the required fields are different, you may list that as dependency instead.

Although it worth asking someone more knowledgeable on React's inner magic, about this case.

coffeeAGNTCY/coffee_agents/corto/exchange/main.py
pregnor: - It's called from `coffeeAGNTCY/coffee_agents/corto/exchange/frontend/src/components/MainArea/Graph/Directory/DirectoryApi.ts` at line 68.
- There is a constant function called above at line 26 that provides the values based on the node which had its agent directory button clicked. And that function maps it to 

```
if label1 == "supervisor agent" || 
    (label2 && label2 == "buyer") {
    return "exchange-supervisor-agent" 
} else if label1 == "grader agent || 
    (label2 && label2 == "sommelier") {
    return "flavor-profile-farm-agent"
} else {
    throw new Error("No valid slug mapping found for node: ${label1} ${label2}")
}
```
- So the frontend injection is covered.
- Backend injection is covered by the line forcing local files and no other files than agent JSONs are displayed from there and even if the wrong JSON is displayed, it's just an inline view to show it, no processing happens so worst case we can be hacked to show the wrong agent record.
- Also IMO backend injection should ideally be covered by a more generic approach.
- Also the backend dockerfile doesn't expose the API server port, the docker compose file does expose it and we also create a service for it in kubernetes, but we don't create external access to the service, so it can only be attacked from inside the cluster.
- For this attack to work the attacker would need to put a malicious .json file there that can be triggered while reading the file in Python and also make an internal API call to the API server with the file name.
- Also we use the exact same thing in Lungo and so far we were okay with it.

Aside from all that, I can lock the API file mapping down to the 2 existing files - I would leave the slug approach because it's better than hard coding.

coffeeAGNTCY/coffee_agents/corto/exchange/frontend/src/components/MainArea/Graph/Directory/OasfRecordModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+          className="relative flex max-h-[70vh] w-[550px] flex-col items-start gap-4 rounded-md bg-node-background p-4 shadow-lg"
+          onClick={handleModalClick}
+          data-modal-content
+        >
+          <button
+            onClick={onClose}
+            className="absolute right-3 top-3 z-10 text-xl leading-none text-node-text-secondary transition-colors hover:text-node-text-primary"
+          >
+            ×
+          </button>
```
delthazor, line 86: This might be intentional, but in my experience quite a lot of webpages treat the backdrop click as a close intent on the modal window, not just the close button/escape key. It is not mandatory though, I've seen plenty of counter-examples. Just making sure which user experience are we aiming for.

pregnor: I think it makes sense to leave it open while the user uses the rest of the UI plus Lungo is using it this way, for now I would favor consistency over intuition here, I would fix it in one go everywhere instead if that's ok.

coffeeAGNTCY/coffee_agents/corto/exchange/frontend/src/components/MainArea/Graph/CustomNode.tsx
pregnor: We discussed to defer this privately.

coffeeAGNTCY/coffee_agents/corto/oasf/agents/exchange-supervisor-agent.json
```diff
@@ -0,0 +1,97 @@
...
+        ]
+      },
+      "name": "integration/a2a"
+    }
+  ],
+  "created_at": "2026-02-02T17:59:12Z",
+  "authors": [
+    "Generated by OASF SDK"
+  ]
+}
```
delthazor: Most of our files have trailing newline at eof. I don't know if it is a purposeful style decision, or just happens to be, but the more files follow the same trend, the better.

coffeeAGNTCY/coffee_agents/corto/oasf/agents/flavor-profile-farm-agent.json
```diff
@@ -0,0 +1,98 @@
...
+        ]
+      },
+      "name": "integration/a2a"
+    }
+  ],
+  "created_at": "2026-02-02T17:59:12Z",
+  "authors": [
+    "Generated by OASF SDK"
+  ]
+}
```
delthazor: See the other comment about trailing newline at eof.

pregnor: Fixed in [d6102aa](https://github.com/agntcy/coffeeAgntcy/pull/472/commits/d6102aa35d49385d5a54fe615bdced982d2d684e)

coffeeAGNTCY/coffee_agents/corto/exchange/main.py
pregnor: Fixed in [81ead5e](https://github.com/agntcy/coffeeAgntcy/pull/472/commits/81ead5ea6e1986df7d8387de1b280c668b3d770f).

coffeeAGNTCY/coffee_agents/corto/oasf/agents/exchange-supervisor-agent.json
pregnor: Fixed in [d6102aa](https://github.com/agntcy/coffeeAgntcy/pull/472/commits/d6102aa35d49385d5a54fe615bdced982d2d684e).

delthazor (approved):
Seems ok to me, you can resolve the convos.

PR #473, opened 2026-03-18
feat(recruiter,ci): Recruiter integration tests in workflows

mihaialexandrescu:
# Description

This PR:
  * adds docker deps to the `pytest` infra setup for `recruiter`
  * adds `pytest` fixtures to install `dirctl` to be used for Recruiter tests (inspired from a Recruiter Dockerfile)
  * cleans up the `OASF_API_VALIDATION_DISABLE` env var from the `.env.example` of Recruiter that wasn't referenced anywhere other than the example and the README  (the other Directory-related vars get set by the docker-compose file to useful values and others get defaulted in the application code)
  * copies `docker_helpers.py` from Lungo to Recruiter
  * other additions made to `conftest.py` try to mimic how the corresponding file works in Lungo
  * updates `recruiter` README

## Issue Link

#411 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [x] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,11 +6,13 @@ on:
     paths:
       - 'coffeeAGNTCY/coffee_agents/corto/**'
       - 'coffeeAGNTCY/coffee_agents/lungo/**'
+      - 'coffeeAGNTCY/coffee_agents/recruiter/**'
       - '!coffeeAGNTCY/coffee_agents/*/deployment/helm/**'
   pull_request:
     paths:
       - 'coffeeAGNTCY/coffee_agents/corto/**'
       - 'coffeeAGNTCY/coffee_agents/lungo/**'
+      - 'coffeeAGNTCY/coffee_agents/recruiter/**'
       - '!coffeeAGNTCY/coffee_agents/*/deployment/helm/**'
       - '.github/workflows/test.yaml'
       - '.github/workflows/test-reusable.yaml'


coffeeAGNTCY/coffee_agents/recruiter/docker/docker-compose.yaml
```diff
@@ -28,11 +30,14 @@ services:
...
-      retries: 60
-      start_period: 15s
+    # # No healthcheck: production image is distroless and does not have wget/curl/nc or shells.
+    # # Note: Zot has a /readyz endpoint (Kubernetes compliant API) since v2.1.6 that can be used for healthchecks.
+    # # Notes for /readyz: https://zotregistry.dev/v2.1.14/general/whats-new/#v216 and https://kubernetes.io/docs/reference/using-api/health-checks/
+    # healthcheck:
+    #   test: curl -f http://localhost:5000/v2/ || exit 1
+    #   interval: 5s
+    #   retries: 60
+    #   start_period: 15s
```
pregnor, line 40: **Question**: If there is a `/readyz` endpoint don't we want to try using that? Is there a reason not to?

Edit: we use this already in the test setup don't we? Why don't we use it here?

coffeeAGNTCY/coffee_agents/recruiter/.gitignore
pregnor, line 1: **Recommended**: IMO we **SHOULD** use the repo root gitignore instead with a full path, because this makes finding ignores more difficult by segregating the ignores. We can do a `/coffeeAGNTCY/coffee_agents/recruiter/bin/**` ignore there (also we should use the double asterisk, because of subfolders, but my knowledge might be outdated and it might be redundant). Or we could even do `**/bin/**` in general on top level.

Edit: I would say `**/bin/**` on top level is the practice I'm used to.

coffeeAGNTCY/coffee_agents/recruiter/test/integration/conftest.py
```diff
@@ -2,19 +2,120 @@
...
+
+    Prepends recruiter/bin to PATH when using the downloaded binary so subprocess calls to `dirctl` work.
+    """
+    if _find_executable_on_path("dirctl"):
+        return
+    if _path_is_executable(LOCAL_DIRCTL):
+        os.environ["PATH"] = f"{BIN_DIR}{os.pathsep}{os.environ.get('PATH', '')}"
+        return
+    _download_dirctl(LOCAL_DIRCTL)
+    os.environ["PATH"] = f"{BIN_DIR}{os.pathsep}{os.environ.get('PATH', '')}"
```
pregnor, line 107: **Recommended**: Under normal circumstances we **SHOULD** simplify this with less control flow jump and logic jumps as well resulting in less cognitive load, but now I'm not forcing this.

```suggestion
    if _find_executable_on_path("dirctl"):
        return
    
    if not _path_is_executable(LOCAL_DIRCTL):
        _download_dirctl(LOCAL_DIRCTL)
    
    os.environ["PATH"] = f"{BIN_DIR}{os.pathsep}{os.environ.get('PATH', '')}"
```

coffeeAGNTCY/coffee_agents/recruiter/test/integration/conftest.py
```diff
@@ -2,19 +2,120 @@
...
+
+
+@pytest.fixture(scope="session")
+def dirctl_path(ensure_dirctl) -> Path:
```
pregnor, line 111: **Required**: We **MUST** add `autouse=true` here to ensure calling according to my findings.

coffeeAGNTCY/coffee_agents/recruiter/test/integration/conftest.py
```diff
@@ -39,6 +140,91 @@ def tracking_loop_factory(*args, **kwargs):
...
+        return
+    _session_docker_torn_down = True
+    print("--- Tearing down session Docker (zot, dir-api-server, etc) ---")
+    down(files)
+
+
+atexit.register(_teardown_session_docker)
+
+# Host-side readiness (compose maps zot 5000 to 5555;)
+ZOT_REGISTRY_READYZ_URL = "http://127.0.0.1:5555/readyz"
```
pregnor: **Recommended**: We **SHOULD** move this to top to the rest of the constants to avoid this being hidden.

coffeeAGNTCY/coffee_agents/recruiter/docker/docker-compose.yaml
mihaialexandrescu: Docker Compose `healthcheck` runs inside the container, not outside.  

> No healthcheck: production image is distroless and does not have wget/curl/nc or shells.

So in the container we have no troubleshooting tools or even the possibility of running `sh` or `bash`.  

> we use this already in the test setup don't we? Why don't we use it here?

Because the tests run on the host and we can check the exposed port from the host at host level (as opposed to within the container).

coffeeAGNTCY/coffee_agents/recruiter/test/integration/conftest.py
mihaialexandrescu: Good idea.  
Done in 68439d683763dcddce4a6633e27bd29a1872d439.

coffeeAGNTCY/coffee_agents/recruiter/test/integration/conftest.py
mihaialexandrescu: You're right.  
Done in 68439d683763dcddce4a6633e27bd29a1872d439.

coffeeAGNTCY/coffee_agents/recruiter/.gitignore
mihaialexandrescu: Done in 19d39e0786d7f79b2939da8876d77614804339d1.

.github/workflows/test.yaml
```diff
@@ -177,3 +206,14 @@ jobs:
...
+  integration-tests-recruiter:
+    needs: subprojects-to-test
+    if: ${{ needs.subprojects-to-test.outputs.test_recruiter == 'true' }}
+    uses: ./.github/workflows/test-reusable.yaml
+    with:
+      project_dir: coffeeAGNTCY/coffee_agents/recruiter # cannot access env context here
+      pip_overrides: ${{ inputs.pip_overrides }}
+      pip_constraints: ${{ inputs.pip_constraints }}
+      docker_overrides: ${{ inputs.docker_overrides }}
+    secrets: inherit
```
delthazor: Most files we have, has a trailing newline eof. Not important, but more files follow the same style, the better.

coffeeAGNTCY/coffee_agents/recruiter/test/integration/conftest.py
```diff
@@ -39,6 +143,88 @@ def tracking_loop_factory(*args, **kwargs):
...
+    to lungo's session slim/nats/otel compose setup.
+    """
+    print("\n--- Setting up session level service integrations ---")
+    down(files)
+    remove_container_if_exists("docker-dir-api-server-1")
+    remove_container_if_exists("docker-zot-1")
+    setup_directory_services()
+    print("--- Session level service setup complete. Tests can now run ---")
+    yield
+    _teardown_session_docker()
```
delthazor, line 205: I may be mistaken, but I think this may not clean up properly, if the test was cancelled midway. In lungo integration conftest, we have a similar scenario, where a try/finally block tries to prevent incomplete teardown for this scenario. It might be irrelevant for recruiter cases, since the setup is much less complex here.

edit: typo

mihaialexandrescu: I don't follow.  

In Lungo in https://github.com/agntcy/coffeeAgntcy/blob/42b47159bcbf470e164e82a59774e7cb0dc1a44a/coffeeAGNTCY/coffee_agents/lungo/tests/integration/conftest.py we have:
```python
_session_docker_torn_down = False

def _teardown_session_docker():
    """Run docker compose down so session containers (slim, nats, grafana, etc.) are stopped. Idempotent."""
    global _session_docker_torn_down
    if _session_docker_torn_down:
        return
    _session_docker_torn_down = True
    print("--- Tearing down session Docker (slim, nats, otel-collector, clickhouse, grafana) ---")
    down(files)

atexit.register(_teardown_session_docker)

# ... 

@pytest.fixture(scope="session", autouse=True)
def orchestrate_session_services():
    print("\n--- Setting up session level service integrations ---")
    down(files)
    remove_container_if_exists("lungo-slim")
    remove_container_if_exists("lungo-nats")
    remove_container_if_exists("lungo-otel-collector")
    remove_container_if_exists("lungo-clickhouse-server")
    remove_container_if_exists("grafana-lungo")
    setup_transports()
    setup_observability()
    setup_identity()
    print("--- Session level service setup complete. Tests can now run ---")
    yield
    try:
        _shutdown_otel_sdk()
    finally:
        _teardown_session_docker()
```

Everything **about teardown** is the same in Recruiter except for the `try` block at the end of `orchestrate_session_services()` which, in Lungo, guards the `_shutdown_otel_sdk()` call, not `_teardown_session_docker()`.  

In Recruiter, since we don't have `otel` shutdown that can fail, I removed the `try` and just called  `_teardown_session_docker()` directly. Is that not equivalent ?

Lastly, isn't `atexit.register(_teardown_session_docker)` supposed to help with it all ?

delthazor: Probably you are right about the ataxia.register, that's why I said I my be mistaken :)
I'm not really an expert on the testing framework, and I often have my fights with it, that's why I highlighted the difference from lungo. But sure, the recruiter does not have the telemetry, so the Otel fail is not in picture here.

You can resolve this, I'm fine with your explanation :)

pregnor: I think this part works the current way.

.github/workflows/test.yaml
mihaialexandrescu: You are right.  
Addressed in 39f56f4d6080ad6b291ce0561f11bbcaac543a4a.

PR #505, opened 2026-04-01
(fix) Auction test flakiness fix for timeout errors

### LINKED ISSUE #495: [Bug]: Timeout messages sometimes reported as permission issue
### Bug Description

When a network message timeout happens, the graph first returns a sensible user-facing error (e.g.: “I encountered an issue retrieving information from the Brazil farm…”). After the reflection node runs, that message can be replaced by a hardcoded 'no permission / verify credentials' string — even though the underlying failure was different.

### Expected Behavior

User facing error message should reflect the real failure mode, so a timeout report when timeout happens, and permission errors, when a permission issue happened.

### Affected Version

v1.0.0

### Steps to Reproduce

Run integration test tests/integration/test_auction.py::test_auction_a2a_timeout_returns_user_visible_error (or equivalent: patch send_a2a_with_retry to raise TransportTimeoutError, POST Brazil inventory prompt to the auction supervisor).

### Checklist

delthazor:
# Description

Add minor error origin and type info into AIMessage kwargs so reflection can more easily determine proper action.
## Issue Link

Fix #495 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [ ] Existing issues have been referenced (where applicable)
- [ ] I have verified this change is not present in other open pull requests

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
@@ -13,6 +13,10 @@ from langgraph.graph import StateGraph, END
 from langgraph.prebuilt import ToolNode
 from ioa_observe.sdk.decorators import agent, graph
 
+from agents.supervisors.auction.graph.a2a_retry import (
+    RemoteAgentNoResponseError,
+    TransportTimeoutError,
+)
 from agents.supervisors.auction.graph.tools import (
     get_farm_yield_inventory,
     get_all_farms_yield_inventory_streaming,
@@ -24,6 +28,32 @@ from agents.supervisors.auction.graph.shared import farm_registry
 from common.llm import get_llm
 
 logger = logging.getLogger("lungo.supervisor.graph")
+
+_SUPERVISOR_OUTCOME_KEY = "supervisor_outcome"
+_OUTCOME_TRANSPORT = "transport"
+_OUTCOME_PERMISSION = "permission"
+_SUPERVISOR_FARM_KEY = "supervisor_farm"
+_SUPERVISOR_OPERATION_KEY = "supervisor_operation"
+
+
+def _caused_by_transport(exc: BaseException) -> bool:
+    """True if exc or its __cause__/__context__ chain includes a transport timeout or no-response."""
+    seen: set[int] = set()
+    stack = [exc]
+    while stack:
+        e = stack.pop()
+        eid = id(e)
+        if eid in seen:
+            continue
+        seen.add(eid)
+        if isinstance(e, (TransportTimeoutError, RemoteAgentNoResponseError)):
+            return True
+        if e.__cause__ is not None:
+            stack.append(e.__cause__)
+        ctx = getattr(e, "__context__", None)
+        if ctx is not None and ctx is not e.__cause__:
+            stack.append(ctx)
+    return False
 
 class NodeStates:
     SUPERVISOR = "exchange_supervisor"
@@ -177,7 +207,6 @@ class ExchangeGraph:
             Review the entire conversation history provided.
 
             Decide whether the user's *original query* has been satisfied by the responses given so far. If the prompt is related to order, please ensure the farm information is included in the final response.
-            For permission issues regarding creating a payment or list transaction, please include which operation failed in the final response.
             If the last message from the AI provides a conclusive answer to the user's request, or if the conversation has reached a natural conclusion, then set 'should_continue' to false.
             Do NOT continue if:
             - The last message from the AI is a final answer to the user's initial request.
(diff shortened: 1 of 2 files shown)


coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
```diff
@@ -175,7 +205,6 @@ class ShouldContinue(BaseModel):
             Review the entire conversation history provided.
 
             Decide whether the user's *original query* has been satisfied by the responses given so far. If the prompt is related to order, please ensure the farm information is included in the final response.
-            For permission issues regarding creating a payment or list transaction, please include which operation failed in the final response.
```
pregnor, line 180: **Question**: What is the reason behind removing this?

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
```diff
@@ -206,23 +235,49 @@ class ShouldContinue(BaseModel):
...
+                    )
+                else:
+                    content_lower = last_msg.content.lower() if isinstance(last_msg.content, str) else ""
+                    for farm in ["colombia", "brazil", "vietnam"]:
+                        if farm in content_lower:
+                            err_msg = (
+                                f"The supervisor agent doesn't have permission to access the {farm.title()} farm. "
+                                "Please verify your access credentials and try again."
+                            )
+                            break
```
pregnor: **Question**: Why are we having different branches for farm_kw and farm in content?
In the end If I see it right we are doing the exact same error, I would really separate the "getting the farm from either kw or content" and "then let's create the error message" parts so we don't duplicate the error message.

I would even go beyond and we should extract the `if values then create error` part to a separate function and the "find me the triggers" part to another one and only wire the two together here, because the farm and operation finding is very similar, could be a reusable parametric function.
Same holds true for operation_kw below.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/auction/test_reflection_supervisor_outcome.py
```diff
@@ -0,0 +1,104 @@
...
+                "Issue details:",
+                "I encountered an issue retrieving information from the Brazil farm",
+                "Brazil",
+                "permission",
+            ],
+        ),
+        (
+            "permission_payment_kw_uses_mcp_template_without_payment_in_content",
+            {"supervisor_outcome": "permission", "supervisor_operation": "payment"},
+            "Tool reported authentication failure (code 403). Please try again later.",
```
pregnor: **Question**: Isn't this error dubious at best? 403 shouldn't be fixed with a temporally shifted retry.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
delthazor: The initial reason was that the reason now can be added programmatically, but on a second thought, you are right to question this. It could be reverted, so the agent tries to include it based on the information available.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/auction/test_reflection_supervisor_outcome.py
delthazor: You are right. In our case the authentication error is usually caused by authentication not being ready yet, so trying again later is making sense this time, but generally it is not a good message for an authentication error.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
delthazor: Makes sense, I'll look into it.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
delthazor: Reverted the prompt line.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
delthazor: Refactored to re-use helper functions for the similar parts.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/auction/test_reflection_supervisor_outcome.py
delthazor: Wording fixed, "retry" phrasing removed.

pregnor (approved):

mihaialexandrescu (approved):

PR #513, opened 2026-04-07
Initial version for in-memory state storage

### LINKED ISSUE #448: [Feat] BLS Implement in-memory workflow-instance state store
Phase 2, core implementation

Implement the store that holds the state defined by the events, used later by the instance/state endpoints (implemented separately), and by the SSE (implemented separately).

- The in-memory state store should be able to receive events defined by the schema and update it's `state` state with the event content.
- If necessary queuing should be added to the insertion point of data, so events can be queued and processed async.
- Whenever the state is updated, the state store should also notify the SSE subscribers, the interface must be prepared for this, but the implementation can be deferred until the SSE ticket.
- Events pushed to subscribers should be the same JSON objects that will be sent as SSE NDJSON lines (schema from #446 ).
- No persistence or replay; workflow-instance lifecycle is “current workflow-instance only” per epic.

delthazor:
# Description

Interface and initial approach to in-memory state storage, and state merge functionality, along with corresponding tests.

## Issue Link

Fixes #448 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


codyhartsook (dismissed):
store looks good to me

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/notifier.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class NotifierProtocol(Protocol):
+    def notify(self, instance_id: str, event: dict) -> None:
```
pregnor: **Question**: Why isn't it `event: schema.types.Event`? (with the accompanying `import schema.types` hence, no committable suggestion)

Same below for the `NoOpNotifier`.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -0,0 +1,254 @@
...
+EVENT_SCHEMA = "event_v1"
+
+_STORE_CLOSED_MSG = "WorkflowInstanceStateStore is closed"
+
+logger = logging.getLogger(__name__)
+
+_DISPATCH_SENTINEL: Final[object] = object()
+
+
+def _touched_instance_ids(event: dict) -> list[str]:
```
pregnor: **Question**: Again, why not `schema.types.Event`?

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -0,0 +1,254 @@
...
+        if not isinstance(wf, dict):
+            continue
+        instances = wf.get("instances")
+        if not isinstance(instances, dict):
+            continue
+        for iid in instances.keys():
+            if isinstance(iid, str) and iid not in seen:
+                seen.add(iid)
+                ordered.append(iid)
+    return ordered
```
pregnor: **Note**: With the hard typing this might be replaced with a list comprehension, probably not, but definitely simplified.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class WorkflowInstanceStateStore:
+    """Validate-then-merge store with optional notifier and per-instance subscribers."""
+
+    def __init__(self, notifier: NotifierProtocol | None = None) -> None:
+        self._notifier = notifier if notifier is not None else NoOpNotifier()
+        self._snapshot: dict = {"workflows": {}}
```
pregnor: **Question**: Why aren't we using the `schema.types.Data` type here?

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -0,0 +1,254 @@
...
+                if timeout is None:
+                    self._dispatch_cv.wait()
+                else:
+                    remaining = end - time.monotonic()
+                    if remaining <= 0:
+                        msg = "Timed out waiting for dispatch queue to drain"
+                        raise TimeoutError(msg)
+                    if not self._dispatch_cv.wait(timeout=remaining):
+                        msg = "Timed out waiting for dispatch queue to drain"
+                        raise TimeoutError(msg)
```
pregnor: **Recommended**: Simplify for readability

```suggestion
                    if remaining <= 0 or not self._dispatch_cv.wait(timeout=remaining):
                        msg = "Timed out waiting for dispatch queue to drain"
                        raise TimeoutError(msg)
```

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
pregnor: **Optional**: `state` **MAY** be a more intuitive name here, because we use state to refer to the compound state made up of the previous state and the event state changes.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def _merge_snapshot(self, event: dict) -> list[str]:
+        self._snapshot = merge_event_data(self._snapshot, event)
+        return _touched_instance_ids(event)
+
+    def _submit_core(self, work: dict) -> None:
+        """Merge then enqueue dispatch. Lock is released before ``put`` so listeners can read state."""
+        with self._state_lock:
+            self._check_open_locked()
+            touched = self._merge_snapshot(work)
```
pregnor: **Note**: I came here from line 254 following down my feeling of weirdness.

I'm not a fan of this kind of a lock usage, TBH. I was wrapping my head around why line 177 is not lock guarded before checking for the usages and seeing it is transitively, but not locally.

My problem is that the store owns the lock, but we don't make the two functions idempotent, because the higher level function handles the lock, so the lower level function cannot be called without explicit locking elsewhere, putting the responsibility on each caller instead of the single callee. 

It's easily fixed, with 2 lines of change, I'm just having a problem of not preventing the occurrence of these kind of things by design.

```suggestion
    def _merge_snapshot(self, event: dict) -> list[str]:
        with self._state_lock:
            self._snapshot = merge_event_data(self._snapshot, event)

        return _touched_instance_ids(event)

    def _submit_core(self, work: dict) -> None:
        """Merge then enqueue dispatch. Lock is released before ``put`` so listeners can read state."""
        self.reject_if_closed()

        touched = self._merge_snapshot(work)
```

(BTW Cursor you implemented self.reject_if_closed() for this exact reason, why didn't you use it here for the compound effect.)

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -0,0 +1,254 @@
...
+        work = copy.deepcopy(event)
+        self._submit_core(work)
+
+    async def submit_event(self, event: dict) -> None:
+        """Validate, merge, then queue dispatch. Serialized with an asyncio lock."""
+        self._reject_if_closed()
+        validate_data_against_schema(event, EVENT_SCHEMA)
+        work = copy.deepcopy(event)
+        async with self._async_lock:
+            self._submit_core(work)
```
pregnor: **Note**: Connected to my other comment at line 184,

If we are being called as an async call we are locking both `self._async_lock` and `self._state_lock` as well.
This is what I meant by the private conversation, I have a hunch that we are not doing the locking and the internal logic/invariant optimally/intuitively.

This will also mean that the synchronization of async calls will happen at this very `async with` because the following section will not be entered concurrently.

And my problem is this will wait for the `self._submit_core(work)` because that's a sync call, then wait for the `merge_event_data(self._snapshot, event)` transitively which is also a sync call thus every event submission is waiting for the entire previous event submission line to be processed before proceeding.

Wouldn't it be better to have a background queue here, and do

```
# pseudo
lock event_queue_lock
event_queue.put(event)
unlock event_queue_lock
```

while another background thread of the store - that is scalable BTW in theory, let's forget Python's limitations for a sec - can run a loop where it gets the items from the event_queue, merges them into the state and then goes to do the next one?

For me it feels like the current solution will force everyone to wait during submission for the processing of other's events, while my proposal decouples the submission from the processing and thus limits the waiting to the enqueuing and handles the processing on the store's side without blocking the event emitters during submission.

This IMO would also have the benefit of limiting the lock to the queue writes (and if Python absolutely requires it then to the async lock for the `async with` calls, I'm not sure, but I think that can be spared) and maybe a single separate lock for the store itself (I still don't get why we need the lifecycle hook and that removal of the worker thread so badly, have a store lock and update the open/closed so other ops are not proceeding anymore), but nothing more.
(Now I may have understood my problem with the locks, I feel like we are not tying locks to data entities, but to functionalities which I don't like until I understand the functionality invariants in the type.)

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+            continue
+        eid = item.get("id")
+        if isinstance(eid, str):
+            out[eid] = copy.deepcopy(item)
+    return out
+
+
+def _apply_topology_items(
+    existing_nodes: dict[str, dict],
+    existing_edges: dict[str, dict],
```
pregnor: **Question**: Again, why wouldn't we use hard types here from `schema.types.event`?

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+    items: list[Any] | None,
+    *,
+    kind: str,
+) -> None:
```
pregnor: **Question**: Why are we using a side-effect method instead of a regular function that returns the new value that can be assigned to the same or a different variable?

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+"""
+
+from __future__ import annotations
+
+import copy
+from typing import Any
+
+
+def _stable_topology_list(by_id: dict[str, dict]) -> list[dict]:
+    return [copy.deepcopy(by_id[k]) for k in sorted(by_id.keys())]
```
pregnor: **Note**: I think sorting by ID might be problematic for example on nodes which are in semantic order with a uuid generated for them.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+    _apply_topology_items(
+        nodes_map, edges_map, delta_topology.get("edges"), kind="edge"
+    )
+    out: dict = {}
+    if nodes_map:
+        out["nodes"] = _stable_topology_list(nodes_map)
+    else:
+        out["nodes"] = []
+    if edges_map:
+        out["edges"] = _stable_topology_list(edges_map)
```
pregnor: **Question**: I was checking why we need the `_stable_topology_list()` function at all, because

> Performing list(d) on a dictionary returns a list of all the keys used in the dictionary, in insertion order (if you want it sorted, just use sorted(d) instead).
> 
> https://docs.python.org/3/tutorial/datastructures.html#dictionaries

Then I saw it here, that if I read it right, at line 95,96 we convert the list of nodes and edges into a map, then we process changes (with a side effect method that we shouldn't do BTW), then we convert the resulting list back to a list of nodes and edges at lines 105 and 109.

First of all, I feel like we don't need the stable function, because map iteration is deterministic on insertion order (see above) and we don't change that if I read it right.

Secondly, IMO there are some redundant special case handlings.
Maybe we could simplify things a bit.

Or not even sure if we gain much by list-map-process-list O(n)+O(m)+O(n) transforming, maybe the (list-)process(-list) O(nm) list-list processing would be sufficient, we won't plan 100 nodes and edges, so < 20000 cycles (vs currently 300 cycles).

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+        if not isinstance(eid, str) or op not in (
+            "create",
+            "read",
+            "update",
+            "delete",
+        ):
+            continue
+        match op:
+            case "create" | "read":
+                bucket[eid] = copy.deepcopy(raw)
```
pregnor: **Required**: We **MUST NOT** change the existing value on read. Read is supposed to be a full representation / get functionality, not a set functionality.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+                if eid not in bucket:
+                    continue
+                target = bucket[eid]
+                for key, val in raw.items():
+                    if key == "id":
+                        continue
+                    if isinstance(val, dict) and isinstance(target.get(key), dict):
+                        target[key].update(val)
+                    else:
+                        target[key] = copy.deepcopy(val)
```
pregnor: **Note**: This is the only part where we should need to process the weak typed dictionary kind if Python doesn't provide field to field update mechanism.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+    """Merge instance ``topology`` using per-item ``operation`` dispatch."""
+    nodes_map = _list_to_map(existing_topology.get("nodes"))
+    edges_map = _list_to_map(existing_topology.get("edges"))
+    _apply_topology_items(
+        nodes_map, edges_map, delta_topology.get("nodes"), kind="node"
+    )
+    _apply_topology_items(
+        nodes_map, edges_map, delta_topology.get("edges"), kind="edge"
+    )
+    out: dict = {}
```
pregnor: **Question**: Why are we starting from an empty result and build it sequentially like with the loop at line 112 and 117?
Instead we could assign it the same hard typed object that we have at `existing_topology`, `_apply_topology_items` could be `_merge_topology_items(existing_topology, delta_topology) -> Topology` instead that we assign to `out` or even `existing_topology` right away.

I also don't like doing the merge by item here, nodes and edges separately, that should happen in 1 atomic step in another function (proposed merge).

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,209 @@
...
+            continue
+        wf_out = out["workflows"].setdefault(wf_key, {})
+        _merge_workflow(wf_out, wf_in)
+
+    for key, val in data_in.items():
+        if key == "workflows":
+            continue
+        out[key] = copy.deepcopy(val)
+
+    return out
```
pregnor: **Note**: IMO with hard types these could be much simplified.

pregnor (commented):
In retrospect, if we had time and availability I would have explored whether we could separate the store into 2 subinterfaces, one for the datastore part and one for the subscribe/dispatch one, because it feels like multiple responsibilities right now mixed, but this is a minor thing right now.

I also haven't checked the tests out yet, because of the expected changes.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/notifier.py
delthazor: For all these questions:
Because the changeset was published before the pydantic model was ready, and I forgot to refactor that after that was merged.
Consider all of these comments as accepted, and fixed in the next update.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
delthazor: I'll definitely simplify it with the new types.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
```diff
@@ -0,0 +1,231 @@
...
+        if isinstance(eid, str):
+            out[eid] = copy.deepcopy(item)
+    return out
+
+
+def _clone_topology_bucket(bucket: dict[str, dict]) -> dict[str, dict]:
+    return {k: copy.deepcopy(v) for k, v in bucket.items()}
+
+
+def _apply_one_topology_item(bucket: dict[str, dict], raw: dict) -> dict[str, dict]:
```
codyhartsook, line 68: Nit (performance, non-blocking): Each mutation branch (CREATE, UPDATE, DELETE) calls _clone_topology_bucket, which deep-copies every entity in the bucket. When an event touches K
   nodes on a topology of N nodes, the sequential fold would produce ~K×N deep copies.                                                                                                   
                  
  For the initial in-memory version I think this is perfectly fine, but if we add more agents / events it could become a hot path. A couple of options for later:

  1. Copy-on-write dict wrapper: only clone entries that are actually modified, pass the rest by reference.
  2. Batch-apply with a single clone: clone the bucket once at the start of _merge_topology_delta_maps, then apply all items mutably on that single copy.

  Not a blocker, just flagging for when this is on a hot path with larger agent graphs.

codyhartsook (approved):
lgtm

pregnor (approved):

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/merge.py
pregnor: I agree we should optimize this later.

https://github.com/agntcy/coffeeAgntcy/issues/530

Issue #449, opened 2026-03-16
[Feat] BLS Implement catalog subAPI endpoints of the agentic workflow API


pregnor:
Phase 2, core implementation

Implement the catalog listing endpoints (pattern, use-case, workflow).
No dependency on workflow-instance store; can ship independently.

- Endpoints: #447 
    - `GET /patterns/` 
    - `GET /use-cases/`
    - `GET /agentic-workflows/`
    - `GET /agentic-workflows/{workflow_name}/`
- No params: return list of use-cases with pattern, use_case id, workflow (list of subobjects with high-level workflow details).
- patterns, use_cases param: filter for entities associated with the specified values.
- Source of catalog and starting topologies is left to implementation (config, code, or existing graph definitions). Align with existing Lungo stack (e.g. supervisors under coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/).

PR #517, opened 2026-04-08
feat: add catalog agentic workflows subAPI endpoints

mihaialexandrescu:
# Description

This PR adds the first implementation (prior to live data store) of the following "catalog"-like API endpoints:
(this table is copied from Issue 447)
| Purpose | Endpoint | Parameters | Response |
|--------|----------|------------|----------|
| List patterns | `GET /patterns/` | (none) | List of patterns |
| List use-cases | `GET /use-cases/` | (none) | List of use-cases |
| List of workflows | `GET /agentic-workflows/` | `patterns: []string` (optional), `use_cases: []string` (optional) | List of available workflows optionally filtered for a pattern+use-case combination. |
| List workflow details (starting topology) | `GET /agentic-workflows/{workflow_name}/` | `topology_only: boolean` (optional)  | List all the available details for a single pattern+use-case=workflow combination with the topology - predefined graph (nodes, edges) for that workflow - with the `topology_only` flag only the starting topology is returned. |

<br/>

The `starting_workflows.json` file is meant to be a seed file with the main distinguishing goal of holding the starting topologies and in this case also provides a temporary source of information for the `agentic-workflows` endpoint (and its children).

The `starting_workflows.json` file was obtained (by an AI agent) from the frontend code (where it was hardcoded). Here are a couple of steps:
  * Names are derived from the title field of each GraphConfig in graphConfigsData.tsx. 
  * Starting topologies are mapped from React Flow Node[]/Edge[] definitions in PUBLISH_SUBSCRIBE_CONFIG, GROUP_COMMUNICATION_CONFIG, and DISCOVERY_CONFIG, translated to the event_v1.json schema format with proper node:// and edge:// URIs, node types (customNode, transportNode, group), edge types (custom, branching), and layer indices inferred from the visual hierarchy. 
  * The Colombia Farm's branching MCP connection (to both Weather and Payment MCP servers) was decomposed into two explicit branching-type edges, since the schema requires one source/target per edge.

The Patterns and UseCases were overriden manually afterwards to match the set of values we decided on offline.

I wanted to run some validation on the starting workflows before accepting them. I imagine we'd want to do that prior to loading them into the data store in the future.  
I am running validation on the contents of the `starting_workflows.json` file against a new jsonschema file that translates to "an array of workflows". I used a new file that depends on a portion of the `event_v1.json` schema file (for the `workflow` type and its dependencies) because I didn't want to validate full `Event` types (I did not feel like they weren't characteristic to the notion of a seed file). I believe that the current implementation of the schema validation package does not allow loading/using schema files that depend on `event_v1.json`. 
This implementation of the `workflows.py` file is unlikely to be long-lived as the Backend Live State epic advances so I went with a localised copy of validation logic that suited it. That validation portion should be removed once the regular schema-based functions can handle more arbitrary sets of chained schema files.

<br/>

I tested manually the API endpoints with curl and the results were the expected ones (including the root path redirect and query parameter filtration for workflows).

## Issue Link

#449 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
delthazor: This is no longer true, update the docstring

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
```diff
@@ -0,0 +1,132 @@
...
+        Filesystem path to the JSON data file.  Defaults to
+        ``starting_workflows.json`` co-located with this module.
+
+    The result is cached process-wide after the first successful load.
+    Callers **must not** mutate the returned structure.
+    """
+    global _cached_workflows
+    if _cached_workflows is None:
+        _cached_workflows = _load_and_validate_starting_workflows_from_file(target)
+    return _cached_workflows
```
delthazor: If cache is built, then target is not used anymore. Calling the function with different target has no effect.
If this is the intended behavior, document it, but I advise against it.
If this is not the intended behavior then:
- if only the default target is meant to be used, then skip the param, and use the path as constant (not recommended)
- if multiple targets are meant to be used, then either key the cache by resolved path or filename, or if it should contain only one file (doesn't seem likely), then clear the cache and repopulate when target differs, and document that behavior

I recommend the keyed cache by path, but that may not line up with the flow you intended, so it is up to you to decide.

delthazor (dismissed):

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/starting_workflows_v1.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/agntcy/coffeeAgntcy/schema/jsonschemas/starting_workflows_v1.json",
+  "title": "Starting Workflows (v1)",
+  "description": "Array of workflow definitions describing the initial topologies for each pattern and use-case. Each element reuses the workflow object type from event_v1.",
+  "type": "array",
+  "minItems": 1,
+  "items": { "$ref": "event_v1.json#/$defs/workflow" }
+}
```
delthazor: I'm still not sure if this solution is the right one even if temporary. Adding a separate file that needs to be manually tracked and versioned just to validate starting topology structure seems like it not worth the cost.

Let's wait for @pregnor 's opinion on this.

mihaialexandrescu: I thought it would be a good reminder that, IMO, we ought to have a validation step between reading the seed workflows (assuming we even want such a thing) and commiting them to whatever structure will hold them. 
 
I am open to fully removing it, especially since it's not the end solution.  

I'll let the discussion form here and I'll take action after that.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
mihaialexandrescu: I chose to have a "target" path parameter mostly for testing purposes (provided I were enclined to add those) so that we're not fully dependent on the global var with the default file path.

I see what you mean about how the function reads and acts when the "cache" is already built. I'll think about changing that.

In spite of the naming, that is not a real cache.

My intention was to temporarily have an in-memory version of a portion of what we would serve once the data store is up and wired.  
I know you were working on the store in parallel. Until that got in, I needed a source of information to feed the API responses.  

The "flow" I imagined (though I have no confirmation if it aligns with short and medium term plans) was that we read the starting workflows/topologies (that we expect to exist when the app boots, without taking any action) from **a single file** (for simplicity) (then maybe we validate what we read) then we write it to the data store that will be the source of truth and source of information after that starting point.   


We could choose to read the "seed" workflows from a folder which would have a file per workflow (spitballing ideas). I don't have indications on whether that's what we want. I chose the version that got me to the point of review quicker with this task: one file with a list of workflows.

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/starting_workflows_v1.json
pregnor: ⏳ (I will get back to this post review.)

pregnor: Why isn't `api.agentic_workflows.WorkflowSummaryResponse` or a `list[api.agentic_workflows.WorkflowSummary]` enough to describe this?

The minItems==1 should be coded to the API endpoint OpenAPI schema if we really need it, but IMO an unbound collection is easier to deal with, especially if filling the type iteratively.
If we introduce this we also need to introduce the python type and add an integrity check which is added complexity for not much gain at this point IMO.

pregnor: Is this because we want to validate the `coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json` content against this?

If so I can let this fly, but I would create a refactor ticket.
If we want to do things like this we should split the event JSON schema into per type files most probably and build a more robust and easier to use type generation and JSON schema object validation pipeline that I did not plan for.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
pregnor, line 1: **Note**: It hurts me to see this JSON file here, at this location among the python package stuff, but I don't have a better location proposal, so I will let this fly I guess.

Hopefully in the long run we can derive these information from a proper backend some day soon.

pregnor, line 1: **Question**: Was there no way to derive this graph programmatically from the langgraph supervisor graphs like 

1. `agents.supervisors.logistics.graph.LogisticGraph.graph`
2. `agents.supervisors.auction.graph.ExchangeGraph.graph`
3. (we don't have anything for this right now, it's more complicated, I need to deep think about this async)

It would be great if we could do this **without** building the graph, maybe refactoring the `build_graph` function to be 2 functions, one for building and one for compiling, the `build_graph` can be these two combined as before (`assemble_graph() ; compile_graph()`), but here at the agentic_workflows API we would only call the `assemble_graph()` part that does not compile and get the LangGraph graph.

Then [`graph.get_graph()`](https://reference.langchain.com/python/langchain-core/runnables/base/Runnable/get_graph) can give us the [`Graph`](https://reference.langchain.com/python/langchain-core/runnables/graph/Graph) that gives a pretty good approximation to nodes and edges.

I don't force this to be implemented this way right now, but I think it would be much more forward facing and I would like to consider this angle whenever we get to a crossroad like this in the future.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
```diff
@@ -0,0 +1,379 @@
+[
+  {
+    "pattern": "Supervisor-worker",
+    "use_case": "Coffee Agency",
+    "name": "Publish Subscribe Coffee Farm Network",
+    "starting_topology": {
+      "nodes": [
+        {
+          "id": "node://1a000001-0001-4000-a001-000000000001",
```
pregnor, line 9: **Note**: This ID system is fine for now, but once the UUID generation is set up for the event emission and propagation this should be updated to dynamically generated ones by the single source UUID generator.

I never intended to have these starting topologies hardcoded like this exactly because of the ID generation.

Maybe we should make this coding happen in Python code instead of raw JSON, what cons do you see to that?
(We can still be visually pleasant with it, but we would have the dynamic UUID generation at least.)

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
```diff
@@ -0,0 +1,379 @@
...
+          "weight": 1.0
+        }
+      ]
+    },
+    "instances": {}
+  },
+  {
+    "pattern": "Supervisor-worker",
+    "use_case": "Coffee Agency",
+    "name": "Publish Subscribe Streaming Coffee Farm Network",
```
pregnor: **Optional**: We **MAY** want to name it a bit more consistent with the existing phrases. Not sure, WDYT?



```suggestion
    "name": "Publish Subscribe Streaming Coffee Auction Network",
```

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
```diff
@@ -0,0 +1,132 @@
...
+    pairs: list[tuple[str, Resource]] = []
+    for name in _REF_SCHEMAS:
+        doc = validation.get_schema(name)
+        rid = doc.get("$id")
+        if rid:
+            pairs.append((rid, Resource.from_contents(doc)))
+    return Registry().with_resources(pairs)
+
+
+def _load_and_validate_starting_workflows_from_file(
```
pregnor: **Question**: What was the reason of building another set of validation tools instead of refactoring the `schema.json_schema.py` to be able to handle this?

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
```diff
@@ -0,0 +1,132 @@
...
+        raise errors.SchemaValidationError(exc.message) from exc
+    except Unresolvable as exc:
+        raise errors.SchemaValidationError(
+            str(exc.__cause__ or exc)
+        ) from exc
+
+    return data
+
+
+_cached_workflows: list[dict[str, Any]] | None = None
```
pregnor: **Recommended**: We **SHOULD** move this to around the top of the file for readability.

Also globals should be capital (keep the underscore for protected stuff).

pregnor: **Required**: We **MUST** use hard typing from the `schema.types` package for the workflow type instead of `Any`.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
```diff
@@ -0,0 +1,132 @@
...
+    global _cached_workflows
+    if _cached_workflows is None:
+        _cached_workflows = _load_and_validate_starting_workflows_from_file(target)
+    return _cached_workflows
+
+
+def clear_cached_workflows() -> None:
+    """Discard the cached workflows (useful in tests or after file edits)."""
+    global _cached_workflows
+    _cached_workflows = None
```
pregnor: **Note**: I do not see the value of using this getter/setter approach, I would have opted for the `_CACHED_WORKFLOWS` being initialized globally at init time and then use it. Clearing is setting it to None, resetting is calling the same load function that we do on startup. I don't feel like losing the target parameter takes anything away.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/agentic_workflows/api/test_patterns_endpoints.py
pregnor, line 1: **Recommended**: We **SHOULD** follow the same table testing approach we did with the schema package and elsewhere (where I generated tests) to be consistent, plus because here the tests are pretty table-compatible IMO.

pregnor (changes requested):

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/starting_workflows_v1.json
mihaialexandrescu: > Is this because we want to validate the coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json content against this?

Yes.  

I will look into using the Pydantic types but I don't want to rely on the `response_model` parameter for this because this initial loading of "workflows" belongs to a different logical flow of information, IMO, than the API response serving flow.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
mihaialexandrescu: I agree it would be great.

I had no idea those existed and still have no idea what they really are or to what they belong logically (other than the name "LangGraph", I asssume(?)).  
That also justifies the "static seed file deriver from current UI (where it is redefined statically too...)" approach you currently see in this PR.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
mihaialexandrescu: That set of validation tools was just merged a very short while ago. We didn't even get to actually use it yet in its envisioned context.  
Refactoring it already to **maybe** **also** use it for smth that that recent code doesn't seem geared towards was smth I was not confident adhered to the vision of that validation "package".  
As other comments on the PR prove, we may not even want to use the jsonschema for that or do this validation at all.  
I prefered to bring it quicker to discussion in the context of the PR.

Technical aspect:   Right now, `json_schema.py` preloads `event_type_v1.json` and takes a single other schema file as an input (which may or may not depend on `event_type_v1.json`). The only other current schema file is `event_v1.json` which does depend on `event_type_v1.json`.  
Overall, this means we can't have a longer (or even different!) chain of schema files depending on each other.

About the refactor you mention:
  1. I was not confident I could do quickly
  2. felt like a significant modification of that package's API/envisioned use that I had not run by anybody and could easily conflict with whatever first use there may be planned for it.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
mihaialexandrescu: They should come from the single source UUID generator. I don't believe we have such a thing currently.  
I think currently, UUID generation is scattered across the codebase with each module independently importing and calling `uuid4()` inline. I don't know if the `uuid` module generates a common global uuid generator/source on its backend. I'd have to look into that.

codyhartsook: what do you think of this approach I've been exploring in the A2A middleware, essentially build deterministic id using uuid5 and agent card name? 

```python
def agent_node_id(card: AgentCard) -> str:
    """Deterministic ``node://UUID`` for an agent, derived from its card name.

    Accepts an ``AgentCard`` so callers don't need to know which field
    drives identity, the same function can be reused anywhere an agent
    node ID is needed (topology builders, starting topology generation,
    validation, etc.).

    .. todo:: The generated IDs must be validated against the event_v1 JSON
       Schema (``NodeId`` pattern) and aligned with the node IDs in the
       catalog's ``starting_workflows.json`` so that runtime topology
       updates match the starting topology nodes.
    """
    return f"node://{_uuid_mod.uuid5(_NODE_NS, f'agent-{card.name}')}"
```

codyhartsook: to @pregnor 's point, this would have to happen in python or the starting workflow json could include AgentCard paths and delay id generation until the workflow is read

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
codyhartsook: I would note that langgraph graph topology is not a representative of the network topology and therefore would suggest tackling this at a later date. Langgraph defines the "business logic" workflows while the `starting_workflows.json` I think should be defining the agent service graph topology and cross network interactions. I am thinking of this similar to open-telemetry, langgraph instr will emit spans but the otel dependency graph is built from cross service interactions

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/starting_workflows_v1.json
mihaialexandrescu: I proposed an updated implementation in https://github.com/agntcy/coffeeAgntcy/commit/3df8bb13d45bcb49b54de5b328789139083625f1 using the Pydantic data models in `coffeeAGNTCY/coffee_agents/lungo/schema/types` and removing that new jsonschema file.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
mihaialexandrescu: I proposed an updated implementation in https://github.com/agntcy/coffeeAgntcy/commit/3df8bb13d45bcb49b54de5b328789139083625f1 using the Pydantic data models in `coffeeAGNTCY/coffee_agents/lungo/schema/types` and removing that new jsonschema file.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
mihaialexandrescu: I proposed an updated implementation in https://github.com/agntcy/coffeeAgntcy/commit/3df8bb13d45bcb49b54de5b328789139083625f1 that uses the Pydantic data models in `lungo/schema/types/`.

mihaialexandrescu: Renamed to `_STARTING_WORKFLOWS` and moved closer to the top of the file in https://github.com/agntcy/coffeeAgntcy/commit/3df8bb13d45bcb49b54de5b328789139083625f1.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
mihaialexandrescu: I proposed a different approach in https://github.com/agntcy/coffeeAgntcy/commit/3df8bb13d45bcb49b54de5b328789139083625f1 that removes the parameter (as indicated) and aims to only run the starting workflow collection during the API server's startup and ensure it only does so once (via a bool + a lock).  
I also removed the clear function. I don't know that we were going to write tests using it and the regular code path wasn't going to use it.

coffeeAGNTCY/coffee_agents/lungo/tests/unit/agentic_workflows/api/test_patterns_endpoints.py
mihaialexandrescu: Agreed.  
I tried following that approach in https://github.com/agntcy/coffeeAgntcy/commit/3df8bb13d45bcb49b54de5b328789139083625f1.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/use_cases.py
```diff
@@ -0,0 +1,14 @@
...
+"""Static catalog data for use-cases.
+
+These are hardcoded until catalog contracts stabilize and a persistent
+store is introduced.
+"""
+
+from __future__ import annotations
+
+USE_CASES: list[str] = [
+    "Coffee Agency",
```
codyhartsook: nit but Agentcy should be Agntcy

mihaialexandrescu: Done in 7463a1641818b3223a1f0bdb05019254a1049e16.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
mihaialexandrescu: 0d48d2ba7da9d0208be315fbe4603aa0ebdee787 is a proposed implementation of this idea based on the discussion Cody and I had offline.

I didn't quite find all the OASF cards for all the agents but I added many of them to the `starting_workflows.json` file in that new `agent_record_uri` field of the `node` type.

codyhartsook: After syncing with Patrik, here’s how I think this should work for initial topology.

The first issue is a semantics mix-up: we’re currently using one identity field for two different jobs.

Stable/semantic identity: “which real agent is this across runs?” - used to render the agent in UI
Runtime/instance identity: “which node object is this in this specific execution graph?”
Those must be separated. Runtime IDs are instance-scoped graph addresses (mainly for edge wiring from my understanding), while stable IDs are what the UI uses to resolve and label actual agents.

A small clarification on label: I think label may have been intended to serve as the stable UI identity in the current contract, but that intent is not very explicit right now. We should make that contract clearer (either by explicitly defining label semantics or adding a dedicated stable identity field).

On `agent_record_uri`: I think adding it was the right change, and we should keep it as the stable identity anchor.
The requested adjustment is to stop deriving node.id from OASF record data (for example via generate_node_id_from_oasf_record(record_data)), as it makes id stable/semantic again (sorry for the turn around).

Suggested direction:

Keep `agent_record_uri` (and optionally a derived agent_stable_id) as semantic identity
Revert and allocate node.id at runtime per instance (scoped by session/workflow context)
Keep label as display/UI contract only once explicitly defined 
Ensure UI mapping is runtime node.id -> stable identity -> display label/name

Given the initial scope is service graph rendering, this keeps graph structure correct while making agent identity explicit and understandable in UI.

mihaialexandrescu: I added a set of updates corresponding to the comment above in 0ce552f42a278335ce2a39926e2cab9e2744ac17.

Note that I did not implement the "agent_stable_id" suggestion because it is unclear for me how the envisioned id would look and how it is supposed to interact with the notion that not all "nodes" are agents and how optional/mandatory we see such a field (schema-wise).

pregnor: PartialNode and PartialEdge objects can take additional properties, so subclassing the type and adding a related field IMO should work, if not then just adding the field dynamically on the PartialType object should work worst case, but try it out, I will come back post review.

Edit: And for now regarding the proposal I'm fine with anything that gets us closer to the finish line. We can have this additional field for Agent type nodes, but not for transport nodes, et cetera.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
```diff
@@ -0,0 +1,164 @@
...
+
+from schema.types import EdgeId, NodeId, TopologyNodeItem, Workflow
+
+
+logger = logging.getLogger(__name__)
+
+_DATA_DIR = Path(__file__).resolve().parent
+_STARTING_WORKFLOWS_FILE = _DATA_DIR / "starting_workflows.json"
+
+_STARTING_WORKFLOWS: list[Workflow] | None = None
```
pregnor: **Note**: If we stored them as a `Dict[workflow.name, workflow]` then we could have taken things out at O(1) in the `/agentic-workflows/{workflow_name}/` workflow endpoint and wouldn't need to iterate through all.
It's a micro-optimization so I will let it fly, but let's look at storage data types from efficiency perspective as well in the future please.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
pregnor: Yeah it looks fine by me for now if it works for Cody's ideas.

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/event_v1.json
```diff
@@ -70,7 +71,7 @@
         {
           "type": "object",
           "description": "Full node data (mainly for init/reset). Used in topology definitions.",
-          "required": ["id", "operation", "type", "label", "size", "layer_index"]
+          "required": ["id", "operation", "type", "label", "size", "layer_index", "agent_record_uri"]
```
pregnor: **Recommended**: IMO we **SHOULD** omit `agnet_record_uri` from the required fields, because for transport nodes it doesn't make sense - or anything beyond agent nodes - we can also omit it from line 64, because at line 66 we declare the type can take additional properties, so if we set the field on the objects it will be accepted by Pydantic as well and we don't have to mandate it for non-agent nodes.
As I wrote on the other thread IMO we can subclass the PartialNode to have a hard defined field if Python supports it - 5m to check - if not we can use it with a custom property as all class objects are dictionaries in Python.

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
```diff
@@ -112,6 +113,7 @@ class Node(BaseModel):
     label: Annotated[str, Field(min_length=1)]
     size: Size
     layer_index: float
+    agent_record_uri: Annotated[str, Field(min_length=1)] | None = None
```
pregnor: **Note**: See my above comments regarding this field.

pregnor (commented):
LGTM in general, couple comments.

codyhartsook (dismissed):
lgtm

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json
mihaialexandrescu: I tried an implementation of these suggestions in 6407a5733880133dccb3916f93c9f9d0e71b4871.  
In there I also tried to incorporate the `stable_agent_id` idea.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py
mihaialexandrescu: I tried an implementation of this suggestion in https://github.com/agntcy/coffeeAgntcy/commit/6407a5733880133dccb3916f93c9f9d0e71b4871 (among other suggestions).

I went for `dict[str, Workflow]` because the key value needs to be hashable and Pydantic `RootModel`s by default are not. They can be if we set `frozen=true`. 
That also has some implications for how we build the WorkflowSummary DTO and overall I thought that if we really want to separate a `WorkflowName` type we can, but I am not convinced of that right now. I'd rather get the review on the rest.

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
```diff
@@ -88,6 +97,34 @@ class Size(BaseModel):
     )
 
 
+_AGENT_SPECIFIC_FIELDS: tuple[str, ...] = ("agent_record_uri", "stable_agent_id")
+
+
+def _reject_agent_specific_extra_fields(instance: BaseModel) -> None:
```
pregnor, line 103: **Note**: I see what was the LLM's point with this, but I really think this is a flawed approach, because we do want to allow other extra fields/properties in node/edge context still, so I would remove this functionality.
Forcing Agent subclass validation should happen through using agent subclass type annotations when creating the data under the nodes. I'm not sure whether it works inherently that way or we need to make it be called explicitly, but I feel like this is locking down the type unnecessarily.
I know the field names won't clash, but still the conceptual approach is not right. If we want to do this, then we should have made the Agent subclass another type in the union in the schema describing PartialNode and Node IMO.

We are just duplicating a bunch of stuff and then wire them specifically so they won't screw things up most of the time, instead of fixing the type hierarchy.

But nobody cares other than me and we are way behind, so I will have to let this through.

pregnor (approved):

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
pregnor: todo(@pregnor): file a ticket for fixing this.

codyhartsook (approved):
lgtm

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
mihaialexandrescu: We discussed it a bit more offline.  
We'll deal with the rest as mentioned above.

Issue #519, opened 2026-04-12
[Bug]: SLIM request() default timeout (6s) is too short for LLM-backed farm agent responses


atripathy86:
### Bug Description

SLIM transport times out before farm agent LLM call completes

The agntcy-app-sdk SLIMTransport.request() has a hardcoded default timeout of 6 seconds. The factory's
  _send_request closure calls transport.request() without passing an explicit timeout, so the default is always
  used. The farm agent's LLM call regularly takes 10–15 seconds, causing the SLIM session to expire before the
  response arrives.

  When this happens, the SDK catches the SlimError.SessionError internally, sets the response to None, and then
  crashes with AttributeError: 'NoneType' object has no attribute 'payload' when trying to decode it. This
  AttributeError has no exception chain link back to the original SessionError, so _is_timeout_error in errors.py
  returns False — the retry loop never fires and the request immediately fails as a 502.

### Expected Behavior

SLIM transport should not time out before farm agent LLM call completes

### Affected Version

v1.0.0

### Steps to Reproduce

Make any valid coffee flavor query. Consistently fails when the LLM response takes longer
  than 6 seconds.

Possible Fix:  Wrap transport.request after creation to change the default timeout to 60s. Longer-term, the SDK should expose a configurable request timeout via create_transport kwargs.

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

mihaialexandrescu:
The initial PRs for this task were https://github.com/agntcy/coffeeAgntcy/pull/520 and https://github.com/agntcy/coffeeAgntcy/pull/579.

Our repository has moved quite a bit in the mean time before we got to review those PRs. I didn't want to make the original contributors go through the hassle of bringing the PRs up to date (and they were tackling the same issues) so I created a different PR (https://github.com/agntcy/coffeeAgntcy/pull/768) with their contribution, solving some merge conflicts, some style differences we now watch for and dealing with some error handling updates in the frontend code.

As I mention in the new PR's description, the credit goes to the initial contributors.

PR #520, opened 2026-04-12
fix(exchange): surface real backend errors, fix SLIM transport timeout

### LINKED ISSUE #518: [Bug]:  parseApiError masks backend errors with generic "Request failed" message
### Bug Description

  parseApiError in exchange/frontend/src/utils/const.ts reads data?.message to extract the error detail, but
  FastAPI serializes HTTPException errors as {"detail": "..."}. Since message is never present, the function
  always falls back to the hardcoded "Request failed" string, hiding the real error (timeout, bad gateway,
  validation failure, etc.) from the user.

### Expected Behavior

Frontend should show real error message. Frontend always shows "Request failed" instead of real error message

### Affected Version

v1.0.0

### Steps to Reproduce

Trigger any backend error (e.g. farm agent down). UI shows "Request failed" regardless of
  actual cause.

Possible Fix: Check data?.detail before data?.message.

### Checklist

atripathy86:
- parse FastAPI's 'detail' field in parseApiError instead of 'message' so real error messages reach the UI instead of 'Request failed'
  - patch SLIMTransport.request() default timeout from 6s to 60s to accommodate LLM response times on the farm agent (~10-15s); the SDK's factory calls transport.request() without an explicit timeout, causing sessions to expire before the farm responds

# Description

#519 - SLIM transport times out before farm agent LLM call completes
#518 - Frontend always shows "Request failed" instead of real error message 

## Issue Link

Fixes #518 and #519

## Type of Change

- [X] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
+++ b/coffeeAGNTCY/coffee_agents/corto/exchange/agent.py
@@ -88,6 +88,24 @@ def _parse_a2a_response(response):
 
 _A2A_MAX_ATTEMPTS = 5
 _A2A_BACKOFF_BASE = 3
+_A2A_REQUEST_TIMEOUT_SECONDS = 60  # LLM calls on the farm can take 10–30s; the SDK default of 6s is too short
+
+
+def _patch_transport_timeout(transport, timeout_seconds: int = _A2A_REQUEST_TIMEOUT_SECONDS):
+    """
+    Override the default point-to-point request timeout on a SLIMTransport.
+
+    The SDK's factory calls transport.request() without an explicit timeout, so
+    the transport falls back to its hardcoded 6-second default — far too short for
+    an LLM-backed farm agent. Wrapping the method here changes the default so every
+    request through this transport uses `timeout_seconds` unless the caller overrides it.
+    """
+    original_request = transport.request
+
+    async def request_with_timeout(recipient, message, timeout=timeout_seconds, **kwargs):
+        return await original_request(recipient, message, timeout=timeout, **kwargs)
+
+    transport.request = request_with_timeout
 
 
 async def _send_a2a_with_retry(client, request):
@@ -191,6 +209,7 @@ class ExchangeAgent:
             endpoint=TRANSPORT_SERVER_ENDPOINT,
             name="default/default/exchange"  # SLIM transport requires a routable name (org/namespace/agent) to build the PyName used for request-reply routing
         )
+        _patch_transport_timeout(transport)
         client = await factory.create_client(
             "A2A",
             agent_topic=a2a_topic,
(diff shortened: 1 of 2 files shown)


mairp:
Hey @atripathy86, nice catch on both the timeout and the frontend `detail` parsing — I hit the exact same SLIM 6s timeout issue independently.

I ended up with a simpler fix that patches the default at the class level instead of wrapping each transport instance. In `exchange/main.py`, at module load:

```python
from agntcy_app_sdk.transport.slim.transport import SLIMTransport

# The SDK's default SLIM request timeout is 6s, too short for LLM round-trips.
SLIMTransport.request.__defaults__ = (30,)
```

This changes the default `timeout` parameter on `request()` for all `SLIMTransport` instances, once, at import time — no wrapper function, no per-request patching. It works because `request(self, recipient, message, timeout=6, **kwargs)` has exactly one defaulted positional arg, so `__defaults__ = (30,)` replaces the `6` with `30`.

**Trade-offs vs your approach:**

- One line instead of ~15 (no wrapper closure, no docstring needed)
- Zero per-request overhead (no new closure allocated on every call)
- Applies globally rather than per-instance, which is fine here since we always want a longer timeout

The `detail` fix in `const.ts` is a separate concern and looks good as-is.

Would you be open to swapping `_patch_transport_timeout` for this one-liner? Happy to push a commit to your branch if that's easier.

atripathy86:
Actually that sounds much better. Let's do it! Can you push the commit in the branch and merge at your convenience. I ended up with this method because of the way I was debugging it.

pregnor (commented):
Thanks for the contribution.

The error message detail propagation seems fine, it is useful, I would keep it.

We have approved mairp's PR regarding the SLIM timeouts, we are working on unblocking the CI checks for easier contributions and once that's unblocked and merged this PR should be rebased onto that.

With that the PR should be good from my perspective.

In the meantime please take a look at the copilot comments, at least resolving the threads will be required before merging.
Edit: also, please sign off on your commits to satisfy the DCO check.

@codyhartsook please also take a look at this PR.

Edit: tracking for the CI unblocking: https://github.com/agntcy/coffeeAgntcy/issues/595

mihaialexandrescu:
Thank you for your contribution!

Our repository has moved quite a bit in the mean time before we got to review this PR. I didn't want to make you go through the hassle of bringing the PR up to date so I created a different PR (https://github.com/agntcy/coffeeAgntcy/pull/768) with your contribution, solving some merge conflicts and some style differences we now watch for and some additional frontend error message handling.
As I mention in that PR's description, the credit goes to you.

mairp:
Appreciate you taking this upstream, @mihaialexandrescu  Same as I said over on #579 — makes sense to consolidate rather than have us rebase two overlapping PRs. Good outcome having the one-liner default-timeout fix and @atripathy86 's frontend detail parsing both land clean in #768. Nice work.

Issue #453, opened 2026-03-16
[Feat] BLS Emit workflow-instance state progress from LangGraph/agents via A2A to state middleware


pregnor:
Phase 4, Emission and documentation

Wire LangGraph execution (e.g. astream_events, graph node callbacks) and/or agents to send custom events (e.g. node start/end, handoff, tool start/end) via A2A to the state middleware so the store and SSE stay in sync with graph execution.

- Build or use adapters that turn LangGraph events into workflow-instance state progress messages conforming to A1’s schema and send them via A2A to the middleware (#452).
- Where events originate: e.g. LangGraph astream_events (v2) in logistics and auction; recruiter’s existing A2A side-channel. Prefer reusing existing A2A patterns where possible.
- Existing NDJSON streaming (e.g. /agent/prompt/stream) remains; this adds the A2A emission path alongside.

Issue #452, opened 2026-03-16
[Feat] BLS A2A middleware for state endpoint


pregnor:
Phase 3, State API, SSE, and A2A middleware

Implement the component that receives custom A2A messages (workflow-instance state progress payloads), validates them against the defined JSON schema, sends update to the in-memory store's endpoint (see #447) for the relevant workflow-instance. All state ingestion goes through this path (no direct agent-to-store calls).

- Incoming A2A messages: validate against schema from #446; reject or log invalid payloads.
- On valid message: update store (write/merge state and/or push event for that workflow-instance via #448’s interface).
- “State agent” or “state topic” and exact A2A transport are left to implementation; epic requires A2A as the single ingestion path.

PR #528, opened 2026-04-22
feat(lungo): emit workflow-instance state progress via A2A middleware (#453)

codyhartsook:
Refs: [#453](https://github.com/agntcy/coffeeAgntcy/issues/453), [#452](https://github.com/agntcy/coffeeAgntcy/issues/452)

## Summary

This PR adds A2A workflow-progress event emission so workflow state stays aligned with supervisor execution across auction, logistics, and recruiter flows.

### What changed

- Bumps `ioa-observe-sdk` and removes the auction `streaming=False` workaround now that the upstream fix has landed.
- Adds A2A event middleware that emits:
  - outbound `create` events when a tool starts an A2A call
  - inbound `update` events as responses arrive
- Attaches `trace_id` and `span_id` to emitted events for downstream correlation.
- Uses OTel span lifecycle to manage in-flight event state, so cleanup happens when the tool-owning span ends.
  - This keeps fan-out, group chat, and recruiter streaming flows in sync without ad-hoc completion counters.
- Adds a workflow registry and tool-call context helpers to map tool calls to workflow definitions.
- Wires the middleware into all supervisor A2A call sites:
  - **Auction**
    - `get_farm_yield_inventory`
    - `get_all_farms_yield_inventory`
    - `get_all_farms_yield_inventory_streaming`
    - `create_order`
    - `get_order_details`
  - **Logistics**
    - `create_order`
    - `create_order_streaming`
  - **Recruiter**
    - `recruit_agents`
    - `evaluate_agent`
    - `DynamicWorkflowAgent._send_a2a_message`
- Threads `ClientCallContext` through `send_a2a_with_retry`.
- Adds `WORKFLOW_API_URL` and `EMIT_WORKFLOW_EVENTS` config flags.

### Sample Output Payload
https://gist.github.com/codyhartsook/5a3e54eec2b694ea8fe5f8a12b452b56

### Automated

- `pytest tests/unit/common/test_a2a_event_middleware.py`
- `pytest tests/integration/test_auction.py::test_auction_create_order_brazil`

### Manual smoke

- **Auction**
  - Verified end-to-end
  - Confirmed workflow events are posted for both single-recipient and fan-out tool calls
  - Confirmed outbound `create` and inbound `update` events share `correlation_id` and `instance_id`
  - Confirmed `get_all_farms_yield_inventory` emits 3 inbound updates for 3 remote farms

- **Logistics**
  - Verified `create_order` and `create_order_streaming`
  - Confirmed group-chat responses stay associated with the owning tool span
  - Verified helpdesk experimental broadcast with `EXPERIMENTAL_FEATURE=true`

- **Recruiter**
  - Verified `recruit_agents`, `evaluate_agent`, and `DynamicWorkflowAgent._send_a2a_message`
  - Confirmed events are attributed to `ON_DEMAND_DISCOVERY`

## Follow-ups (tracked separately)

- [#531](https://github.com/agntcy/coffeeAgntcy/issues/531) — Emit workflow-instance topology events for MCP calls (parity with this A2A middleware). Scoped as a follow-up: refactor the middleware into a shared transport-neutral core with A2A and MCP adapters so farm → MCP edges show up in the same workflow-instance topology. Not in this PR.

## Remaining work

- Add end-to-end integration coverage for logistics and recruiter workflow-event flows
- Call `HttpEventSink.aclose()` during supervisor shutdown


delthazor (commented):
I may be not fully up to date yet, with everything involved, but it seems fine.

Although it might worth a double-check if parallel A2A calls (tool calls) during the same graph step can keep the correct workflow context (it seems correct, but I'm not entirely sure)

codyhartsook:
> I may be not fully up to date yet, with everything involved, but it seems fine.
> 
> Although it might worth a double-check if parallel A2A calls (tool calls) during the same graph step can keep the correct workflow context (it seems correct, but I'm not entirely sure)

Will try to confirm this, workflow instance is based on trace session id so should be but will confirm.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
```diff
@@ -531,3 +543,25 @@ async def get_order_details(order_id: str) -> str:
...
+_event_interceptor = EventEmittingInterceptor(
+    caller_card=AUCTION_SUPERVISOR_CARD,
+    workflow_resolver=_workflow_resolver.resolve,
+    verbose=True,
+)
+_event_consumer = make_event_emitting_consumer(
+    caller_card=AUCTION_SUPERVISOR_CARD,
+    workflow_resolver=_workflow_resolver.resolve,
+    verbose=True,
+)
```
pregnor: **Recommended**: Globals should be at the top for visibility reasons.

Same at coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/logistics/graph/tools.py.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware.py
```diff
@@ -0,0 +1,891 @@
...
+# ---------------------------------------------------------------------------
+# Tool → workflow identity mapping
+# ---------------------------------------------------------------------------
+
+@dataclass(frozen=True)
+class WorkflowIdentity:
+    """Workflow identity values sourced from the workflow catalog."""
+    workflow_name: str
+    pattern: str
+    use_case: str
```
pregnor: **Note (out of PR scope)**: In the long run, I would rather use a schema type for this, such as `WorkflowMetadata`

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware.py
```diff
@@ -0,0 +1,891 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Low-level node / edge factories
+# ---------------------------------------------------------------------------
+
+_DEFAULT_NODE_SIZE = Size(width=1.0, height=1.0)
+
+
+def _make_node(
```
pregnor: **Recommended (out of PR scope)**: In the long run IMO these functions should be added to the types generated from the schema (schema/types/event.py) in order to have them generically placed and have a proper init function that can accept the dynamic init values.

Edit: maybe not the specific values, but the generic init function at least. If we can properly define the major things in schema then even the specifics.

pregnor (dismissed):
High level I think I like the approach, especially regarding the data model.

There are a couple patterns I didn't like (some organization stuff), but those are not hard blockers and I want to progress, so we can always refactor later.

Also I found it a bit AI-smelly (and because of that it felt a bit convoluted at points), but I don't want to nitpick on low level stuff until the linters are set up properly to enforce.

Otherwise LGTM, make sure to manually test it before merging.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
codyhartsook: Will address clarity in a small refactor and add a design doc + session event output for sample runs

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/middleware.py
```diff
@@ -0,0 +1,677 @@
...
+				caller_node,
+				transport_node,
+				operation=Operation.UPDATE,
+				allocator=allocator,
+			),
+		],
+	)
+
+
+class EventEmittingInterceptor(ClientCallInterceptor):
```
delthazor, line 387: This is both defined in `common/a2a_event_middleware/middleware.py` and in `common/a2a_event_middleware.py` and I don't really see why... The two files seem to overlap at other places too. What is the reason in defining a single file module and a full directory package with the same name and having similar content? Is one of them just a leftover, or is there something else I missed?

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
pregnor: This still stands.

pregnor (dismissed):
Organization is better now.

I still have a couple remarks, but overall I think it's good enough to be merged, just check the duplication Peter mentioned.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/middleware.py
codyhartsook: should have been removed, fixed now

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
codyhartsook: can not simply be moved to the top due to order of events, decorators need to be registered. Adding Lazy accessor, deferring construction of middleware until the actual A2A call.

delthazor (dismissed):
lgtm

pregnor (dismissed):
LGTM, give it one last manual test after all the changes

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/tools.py
pregnor: Sorry about that, I'm fine with this, thanks for the improvements.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware.py
pregnor: Created a ticket for this to not forget it: https://github.com/agntcy/coffeeAgntcy/issues/546

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware.py
pregnor: Related ticket: https://github.com/agntcy/coffeeAgntcy/issues/546

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/inflight.py
```diff
@@ -0,0 +1,272 @@
...
+		if span_context is None:
+			return
+		trace_id = span_context.trace_id
+		ending_span_id = span_context.span_id
+		state = in_flight.get(trace_id)
+		if state is None:
+			return
+		if state.owner_span_id != ending_span_id:
+			return
+		in_flight.pop(trace_id, None)
```
delthazor, line 155: I'm not sure, if I see the whole flow here... What happens if an upsert took the in_flight_lock, when this pop is called? Is it even a possible scenario during the cleanup?

delthazor (commented):
Seems solid.

pregnor (approved):

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/inflight.py
codyhartsook: Good question, I dug into this a bit. The short answer is no, it's safe under the current concurrency/A2A model in the supervisors:

asyncio.Lock only serializes concurrent upsert tasks on the event loop. It does not (and can't) fence against on_end, since asyncio.Lock has no thread semantics. The reason on_end is safe without the lock is:

1. on_end runs synchronously on the event loop thread. ioa_observe wires a span processor whose on_end is invoked inline when the span's with block exits, which happens on the same thread running our async tool/interceptor code. There's no await inside on_end, so no other coroutine can interleave between the get on line 150 and the pop on line 155.
2. The owner_span_id guard prevents popping a re-keyed entry. A concurrent upsert that re-touches the same trace_id preserves owner_span_id from the existing entry (line 227), so even in a hypothetical cross-thread scenario the entry being popped is still the legitimate owner's. Any upsert arriving after the owner's span ends is effectively a use-after-end and would produce stale state regardless of locking.

Would you recommend we harden this now? Today it's safe because A2A interceptor calls run on the event loop. If we ever switched to thread-based A2A client calls, asyncio.Lock wouldn't serialize against on_end and the race would come back.

delthazor: I understand your reasoning, and I don't think that switch will happen anytime soon, so no, if this is the case, I would leave it this way. Thanks for clarifying.

delthazor (approved):
lgtm

pregnor:
I merged this to progress with the release in the absence of Cody's.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/workflow_catalog.py
```diff
@@ -0,0 +1,110 @@
...
+logger = logging.getLogger("lungo.common.workflow_catalog")
+
+
+@dataclass(frozen=True)
+class WorkflowMetadata:
+	"""Workflow descriptive metadata sourced from the workflow catalog."""
+
+	workflow_name: str
+	pattern: str
+	use_case: str
```
mihaialexandrescu, line 30: Instead of re-defining it here, shouldn't this be taken from the `DTOs` (where very soon we will be defining them from the OpenAPI spec at `coffeeAGNTCY/coffee_agents/lungo/schema/openapi` using a Skill):
https://github.com/agntcy/coffeeAgntcy/blob/2109ed2f4d65e19453828bcbf5df153d67acf7fa/coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/dtos.py#L44-L50

Example source:
https://github.com/agntcy/coffeeAgntcy/blob/2109ed2f4d65e19453828bcbf5df153d67acf7fa/coffeeAGNTCY/coffee_agents/lungo/schema/openapi/components/schemas.yaml#L67-L83

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/workflow_catalog.py
```diff
@@ -0,0 +1,110 @@
...
+		raise FileNotFoundError(f"Starting workflows data file not found: {path}")
+
+	try:
+		with open(path, encoding="utf-8") as fh:
+			data = json.load(fh)
+	except json.JSONDecodeError as exc:
+		raise RuntimeError(f"Failed to decode {path}: {exc}") from exc
+
+	if not isinstance(data, list):
+		raise RuntimeError(f"Expected a JSON array in {path}, got {type(data).__name__}")
```
mihaialexandrescu, line 61: Isn't this part doing what `coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py` already does ?
The result can already be accessed via
https://github.com/agntcy/coffeeAgntcy/blob/2109ed2f4d65e19453828bcbf5df153d67acf7fa/coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/workflows.py#L177-L182
and then re-shaped into the smaller type (because I see we are not returning a full dict of `Workflow`s).  
As a general advantage, if I understand things correctly: that `get_workflows` function I mentioned now returns a dict from memory but the intention over time is to have that retrieve information from the Store. Using the `starting_workflows.json` again in this file here defeats the purpose of hiding the implementation from `workflows.py`.



As an aside, isn't even that summarization what's already being done in the code for the response to the `/agentic-workflows/` endpoint ?
https://github.com/agntcy/coffeeAgntcy/blob/2109ed2f4d65e19453828bcbf5df153d67acf7fa/coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py#L137-L166
Maybe we make that into a separate function there (dict of `Workflow`s -> dict of `WorkflowSummary`) and then reuse it in both places ?

Issue #451, opened 2026-03-16
[Feat] BLS Implement SSE endpoints – workflow instance stream as NDJSON from event/state schema


pregnor:
Phase 3, State API, SSE, and A2A middleware

Implement the SSE endpoints that streams events/state deltas for that workflow instance as NDJSON, with each line conforming to the same A2A message schema defined for the workflow instance event/state.

- Endpoints: #447 
    - `POST /agentic-workflows{workflow_name}/instances/{workflow_instance_id}/events/`
    - `GET /agentic-workflows/{workflow_name}/instances/{workflow_instance_id}/events/stream`
- Subscribe to the workflow-instance’s event stream from the store (#448). Each item pushed to the state store is emitted as one NDJSON line towards the SSE subscribers. If the state store has not implemented the SSE endpoint data propagation on incoming events, it should be implemented here.
- Content of each line must conform to the JSON schema from #446 so frontend and backend share one contract for both A2A and SSE.
- At least one SSE endpoint per workflow-instance; connection lifecycle and auth are out of scope unless the epic specifies them.

PR #535, opened 2026-04-28
Implement SSE endpoints

delthazor:
# Description

Instantiate workflow instance storage and handle SSE for subscribed listeners.

## Issue Link

Implements #451 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
@@ -49,25 +49,23 @@ WORKFLOW_INSTANCE_SSE_QUEUE_HIGH_WATER = max(
 )
 
 
-def workflow_instance_event_to_sse_bytes(event: Event) -> bytes:
-    """One SSE message: a single ``data:`` line with compact ``event_v1`` JSON."""
-    line = json.dumps(
-        event.model_dump(mode="json", exclude_none=True),
-        separators=(",", ":"),
-    )
-    return f"data: {line}\n\n".encode()
+def workflow_instance_event_to_sse_frame(event: Event) -> str:
+    """One SSE message: a single ``data:`` line with compact ``event_v1`` JSON (UTF-8 text)."""
+    payload = event.model_dump(mode="json", exclude_none=True)
+    line = json.dumps(payload, separators=(",", ":"))
+    return f"data: {line}\n\n"
 
 
 def enqueue_workflow_instance_sse_queue_chunk(
-    queue: asyncio.Queue[bytes],
-    chunk: bytes,
+    queue: asyncio.Queue[str],
+    chunk: str,
     *,
     high_water: int,
 ) -> None:
-    """Append ``chunk`` to ``queue``; at/above ``high_water`` backlog, drop oldest first.
+    """Append ``chunk`` (SSE frame text, typically JSON ``data:`` lines) to ``queue``.
 
-    If ``put_nowait`` still raises :class:`asyncio.QueueFull`, drop oldest and retry
-    until the chunk is queued.
+    At/above ``high_water`` backlog, drop oldest first. If ``put_nowait`` still raises
+    :class:`asyncio.QueueFull`, drop oldest and retry until the chunk is queued.
     """
     if queue.qsize() >= high_water and not queue.empty():
         try:
@@ -290,7 +288,7 @@ def create_agentic_workflows_router() -> APIRouter:
         """GET SSE stream; each message is one ``data:`` line of compact ``event_v1`` JSON."""
         store = _workflow_instance_store(request)
         canonical_id = instance_id_from_uuid(workflow_instance_id).root
-        queue: asyncio.Queue[bytes] = asyncio.Queue(
+        queue: asyncio.Queue[str] = asyncio.Queue(
             maxsize=WORKFLOW_INSTANCE_SSE_QUEUE_MAXSIZE
         )
         loop = asyncio.get_running_loop()
@@ -301,7 +299,7 @@ def create_agentic_workflows_router() -> APIRouter:
             wf_block = ev.data.workflows[workflow_name]
             if canonical_id not in wf_block.instances:
                 return
-            chunk = workflow_instance_event_to_sse_bytes(ev)
+            chunk = workflow_instance_event_to_sse_frame(ev)
 
             def _enqueue() -> None:
                 enqueue_workflow_instance_sse_queue_chunk(
@@ -320,8 +318,8 @@ def create_agentic_workflows_router() -> APIRouter:
             yield b":\n\n"
             try:
                 while True:
-                    chunk = await queue.get()
-                    yield chunk
+                    frame = await queue.get()
+                    yield frame.encode("utf-8")
             finally:
                 unsub()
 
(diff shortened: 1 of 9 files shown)


coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/server.py
```diff
@@ -20,6 +26,20 @@
 logger = logging.getLogger("lungo.agentic_workflows.server")
 
 
+@asynccontextmanager
+async def _close_workflow_instance_store_on_shutdown(app: FastAPI) -> AsyncIterator[None]:
```
mihaialexandrescu: **Suggestion , very optional**:

Since the FastAPI constructor takes a single `lifespan` function, I think I would preemptively write it to have a `lifespan` function and call inside that other functions that take the specific actions, like in their example at https://fastapi.tiangolo.com/advanced/events/#lifespan.

```python
async def _close_workflow_instance_store_on_shutdown():
   # close store
   # ... ...

@asynccontextmanager
async def lifespan(app: FastAPI):
   # do stuff for startup

    yield

    # do stuff for cleanup
    _close_workflow_instance_store_on_shutdown(...)


app = FastAPI(lifespan=lifespan)
```

mihaialexandrescu (dismissed):

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/server.py
delthazor: done

mihaialexandrescu (dismissed):

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
```diff
@@ -24,12 +28,72 @@
...
+)
+
+
+def workflow_instance_event_to_sse_bytes(event: Event) -> bytes:
+    """One SSE message: a single ``data:`` line with compact ``event_v1`` JSON."""
+    line = json.dumps(
+        event.model_dump(mode="json", exclude_none=True),
+        separators=(",", ":"),
+    )
+    return f"data: {line}\n\n".encode()
```
pregnor: **Question**: Any particular reason we encode in this format (YAML??, Python representation??) and use bytes instead of string?

My problem is we are mixing JSON and nonJSON encoding and I don't see why or whether it has any upside instead of storing proper JSON objects as strings for example (or bytes, we can lament on that).

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
```diff
@@ -24,12 +28,72 @@
...
+            pass
+    while True:
+        try:
+            queue.put_nowait(chunk)
+            return
+        except asyncio.QueueFull:
+            try:
+                queue.get_nowait()
+            except asyncio.QueueEmpty:
+                pass
```
pregnor, line 83: **Question**: Why are we adding high water - BTW what's up with everyone's sudden sailor lingo interest - check if there is a full queue check?
What value is added by setting the queue to start dropping at 90 when the queue size is 100 instead of setting the queue size to 90 instead and being more direct and intuitive.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
```diff
@@ -196,9 +286,53 @@ async def stream_workflow_instance_events(
...
+
+        return StreamingResponse(
+            event_bytes(),
+            media_type="text/event-stream",
+            headers={
+                "Cache-Control": "no-store",
+                "Connection": "keep-alive",
+                "X-Accel-Buffering": "no",
+            },
+        )
```
pregnor, line 334: **Note**: At some point I want to revisit this approach, I don't have a suggestion yet, but something feels off to me.
No functional issue, rather just approach-wise.

Edit: maybe that we are trying to use a queue like a Go channel here and I'm not sure it's the most straightforwards way of doing this in Python.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/errors.py
```diff
@@ -0,0 +1,15 @@
...
+from __future__ import annotations
+
+_WORKFLOW_INSTANCE_STORE_CLOSED_MSG = "WorkflowInstanceStateStore is closed"
+
+
+class WorkflowInstanceStoreClosedError(RuntimeError):
+    """Raised when the store is closed or shutting down and the operation cannot proceed."""
+
+    def __init__(self, message: str = _WORKFLOW_INSTANCE_STORE_CLOSED_MSG) -> None:
+        super().__init__(message)
```
pregnor: **Note**: Eh, I would be careful building such a generic type and then allowing overriding the message, I would rather hardcode it in the init, because I don't see a benign use-case where we would want to override this specific error's message, on the other hand it could trick us into activating error branches if someone manipulated this. Or it could throw us off of a debugging course as well.
But feel free to enlighten me if I miss something here.

pregnor (dismissed):

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
delthazor: No particular reason, just som AI slop I glossed over... (shame on me...)

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
delthazor: I watched some pirate movies lately....
Anyways: the idea is to mitigate the chance of the queue getting full. I don't want to drop messages from the get-go, when we don't have to.
But if the queue is getting close to being full, then older, unprocessed messages could be dropped IF there is a newer version of the same payload. This doesn't mean the queue cannot get any "fuller" if completely new payloads are added, but given the sufficient queue size it just becomes quite unlikely. Of course 100 and 90 are both quite big for this, at least compared to our current setup of possible workflows, but the main point I not the size, but the safety mechanism that tries to save the queue from being totally full.
Honestly, I don't think that there would be any more than 10 messages queued up at any given time, without a receiver processing them up, so this feature idea is more of a showcase than a real need. However, I still think, if such case happens, it is better to start dropping messages, that can be (relatively) safely dropped from the queue before it gets full, then dropping everything after it is already full, but without dropping messages from the start.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
delthazor: Sure thing. I think this is fairly common in python, but we can revisit anytime

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/errors.py
delthazor: Ok, then I hardcode the message instead.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
pregnor: My bad, this is how the text/event-stream objects are expected in the streaming response.

pregnor (approved):

mihaialexandrescu (approved):

PR #536, opened 2026-04-30
feat(lungo): add Skill for jsonschema to pydanticv2 types

mihaialexandrescu:
# Description

This PR aims to add a Skill specific to the Lungo subproject that can take the JSONSchema files and generate Pydantic v2 models from them.

The skill is more complicated than initially thought because there are paradigm differences between JSONSchema (and how it composes types) vs Python and Pydantic types/models (and how they treat and propagate settings like `extra` and type composition in general).   
Also with that in mind, the examples from `mapping-rules.md` should make the generation more reliable and closer to deterministic.

The first two commits are small fixes I brought along the way because the coding agents kept getting fixated on those things when trying to do the iterations that would eventually be the basis for the generation of the Skill itself. They are otherwise not related to the main goal of the ticket.

## Issue Link

#510 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.agents/skills/jsonschema-to-pydantic-lungo/mapping-rules.md
+++ b/.agents/skills/jsonschema-to-pydantic-lungo/mapping-rules.md
@@ -1,0 +1,245 @@
+# Mapping rules — worked examples
+
+Worked examples for the rules in [`SKILL.md`](SKILL.md) that are easy to get wrong. Each section shows a JSON Schema fragment and the Pydantic v2 emission for it.
+
+## §A — Embed in-place constraints in the type, not in a validator
+
+### Pattern on a string
+
+Schema:
+
+```json
+"event_id": {
+  "type": "string",
+  "pattern": "^event://[0-9a-fA-F]{8}-...$",
+  "description": "Unique id for an event message."
+}
+```
+
+Emit:
+
+```python
+_EVENT_ID_REGEX = r"^event://[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+
+class EventId(RootModel[str]):
+    """Generated from ``$defs.event_id``: <description>."""
+
+    root: Annotated[
+        str,
+        Field(pattern=_EVENT_ID_REGEX, description="Unique id for an event message."),
+    ]
+```
+
+### `propertyNames` on a dict-shaped property
+
+Schema:
+
+```json
+"instances": {
+  "type": "object",
+  "propertyNames": { "$ref": "#/$defs/instance_id" },
+  "additionalProperties": { "$ref": "#/$defs/workflow_instance" }
+}
+```
+
+Emit (inside the parent class):
+
+```python
+instances: dict[
+    Annotated[str, Field(pattern=_INSTANCE_ID_REGEX)], WorkflowInstance
+]
+```
+
+Do not emit a separate `@field_validator` that loops the keys checking the pattern — the dict-key annotation is enforced by Pydantic on every validation pass.
+
+## §C — `allOf [partial, {required: [...]}]` → standalone full class
+
+Schema:
+
+```json
+"partial_regular_node": {
+  "type": "object",
+  "required": ["id", "operation"],
+  "properties": {
+    "id":          { "$ref": "#/$defs/node_id" },
+    "operation":   { "$ref": "#/$defs/operation" },
+    "type":        { "type": "string", "minLength": 1 },
+    "label":       { "type": "string", "minLength": 1 },
+    "size":        { "$ref": "#/$defs/size" },
+    "layer_index": { "type": "number", "default": 0 }
+  },
+  "additionalProperties": true
+},
+"regular_node": {
+  "allOf": [
+    { "$ref": "#/$defs/partial_regular_node" },
+    {
+      "type": "object",
+      "required": ["id", "operation", "type", "label", "size", "layer_index"]
+    }
+  ]
+}
+```
+
+Emit two unrelated classes (no inheritance):
+
+```python
+class PartialRegularNode(BaseModel):
+    model_config = ConfigDict(extra="allow")
+
+    id: NodeId
+    operation: Operation
+    type: Annotated[str, Field(min_length=1)] | None = None
+    label: Annotated[str, Field(min_length=1)] | None = None
+    size: Size | None = None
+    layer_index: float = 0
+
+
+class RegularNode(BaseModel):  # NOT (PartialRegularNode)
+    model_config = ConfigDict(extra="allow")
+
+    id: NodeId
+    operation: Operation
+    type: Annotated[str, Field(min_length=1)]
+    label: Annotated[str, Field(min_length=1)]
+    size: Size
+    layer_index: float
+```
+
+Why no subclassing: `type: ... | None = None` from `PartialRegularNode` would leak into `RegularNode` and `RegularNode(id=..., operation=...)` would silently validate.
+
+## §D — `anyOf` discriminated by sibling-key presence
+
+Schema (the canonical lungo case for `partial_node` / `node`). The agent-flavoured shapes are themselves first-class `$defs` (`partial_agent_node`, `agent_node`), so the `anyOf` branches reference them by name rather than composing them inline:
+
+```json
+"partial_node_agent_extension": {
+  "type": "object",
+  "required": ["agent_record_uri"],
+  "properties": {
+    "agent_record_uri": { "type": "string", "minLength": 1 },
+    "stable_agent_id":  { "$ref": "#/$defs/stable_agent_id" }
+  }
+},
+"node_agent_extension": {
+  "allOf": [
+    { "$ref": "#/$defs/partial_node_agent_extension" },
+    { "type": "object", "required": ["agent_record_uri", "stable_agent_id"] }
+  ]
+},
+"partial_agent_node": {
+  "allOf": [
+    { "$ref": "#/$defs/partial_regular_node" },
+    { "$ref": "#/$defs/partial_node_agent_extension" }
+  ]
+},
+"agent_node": {
+  "allOf": [
+    { "$ref": "#/$defs/regular_node" },
+    { "$ref": "#/$defs/node_agent_extension" }
+  ]
+},
+"partial_node": {
+  "anyOf": [
+    {
+      "allOf": [
+        { "$ref": "#/$defs/partial_regular_node" },
+        { "not": { "anyOf": [
+          { "required": ["agent_record_uri"] },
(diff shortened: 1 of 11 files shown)


mihaialexandrescu:
### Explanation about the use of Pydantic Union Discriminator

Two example nodes:

* RegularNode (fake name)

```json
{
  "id": "node://6ae354ec-3975-4c6e-bf1c-c3d5c4c61a4c",
  "operation": "read",
  "type": "transportNode",
  "label": "Transport",
  "size": {
    "width": 1.0,
    "height": 1.0
  },
  "layer_index": 1.0
}  
```
* AgentNode

```json
{
  "id": "node://5be42fef-080b-4af7-a82e-c74a152e3453",
  "operation": "read",
  "type": "customNode",
  "label": "Auction Agent",
  "size": {
    "width": 1.0,
    "height": 1.0
  },
  "layer_index": 0.0,
  "agent_record_uri": "../../agents/supervisors/auction/oasf/agents/auction-supervisor-agent.json",
  "stable_agent_id": "agent://470de4a0-a6bc-5e1b-82dd-e41de2af3f85"
}  
```
Mistakes we could introduce in an AgentNode:

* having "stable_agent_id" but no "agent_record_uri"
* malformed "agent_record_uri" (e.g. "https://../../agents/..." like a crossover between a filesystem path and a web URL)
* etc

As long as we have different types (`AgentNode`, `RegularNode`), we will have something like this in Python/Pydantic: (using fake names but inspired from `TopologyNodeItem`)

```python
NodeItem = Union[AgentNode, RegularNode]  
```
That would read like "A NodeItem can conform to the rules of either an AgentNode or a RegularNode".

Now let's assume that a RegularNode (or even both) have `additionalProperties: true` (or the equivalent in Pydantic of `model_config = ConfigDict(extra="allow")`).
Take the example of an AgentNode that has a "stable_agent_id" but not an "agent_record_uri". Such a node should fail that `Union` validation. But it doesn't. It fails the `AgentNode` validation but it matches and passes the `RegularNode` validation due to the `extra="allow"` and the other fields being fine.

That for us is/should be no good.

With the solution around Discriminator functions for Union types, we can address this "unwanted fallthrough" logic.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -286,7 +287,7 @@ def __init__(self, notifier: NotifierProtocol | None = None) -> None:
         self._dispatch = _DispatchHub(self._state_lock, n)
         self._merge = _MergeCoordinator(
             self._state_lock,
-            Data(),
+            Data(workflows=dict[str, Workflow]()),
             self._after_merge_enqueue_dispatch,
         )
```
delthazor, line 292: Please double check if all call sites (especially in tests) are updated accordingly. It might not fail, but can lead to ambiguity later on. (I think the workflow instance state endpoints test might still use the wrong Data() call)

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
```diff
@@ -31,6 +31,7 @@
 import copy
 import logging
 import queue
+from schema.types.event import Workflow
```
delthazor: I don't think we have any linter rules for import ordering that would be broken here, but this still looks a bit off. If you update the PR, please find a better place for this in the import list. 
(If no other issues needs a PR update, then you can leave it, it is not that important)

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
```diff
@@ -58,81 +95,136 @@ class CorrelationId(RootModel[str]):
...
+class Size(BaseModel):
+    """Generated from ``$defs.size``: relative layout size vs other nodes.
+    Used in topology drawing and layout algorithms."""
 
+    model_config = ConfigDict(extra="forbid")
 
-def _reject_agent_specific_extra_fields(instance: BaseModel) -> None:
-    """Reject agent-specific keys if they sneak in via ``extra="allow"``.
+    width: float = Field(default=1.0)
+    height: float = Field(default=1.0)
```
delthazor, line 209: Previous Size class maintained description for the values. I know that they are not very important, but I don't know, if they were removed by accident, or on purpose. I think the jsonschema still defines them, so there is some ambiguity there.

.agents/skills/jsonschema-to-pydantic-lungo/mapping-rules.md
```diff
@@ -0,0 +1,245 @@
+# Mapping rules — worked examples
+
+Worked examples for the rules in [`SKILL.md`](SKILL.md) that are easy to get wrong. Each section shows a JSON Schema fragment and the Pydantic v2 emission for it.
+
+## §A — Embed in-place constraints in the type, not in a validator
+
+### Pattern on a string
+
+Schema:
```
pregnor, line 9: **Question**: Is it not possible to reference these schemas by the JSON schema or does it work better this way?

I'm just asking, because this seems a bit brittle to changes to the event schema.

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/event_v1.json
```diff
@@ -64,16 +69,87 @@
       },
       "additionalProperties": true
     },
-    "node": {
+    "regular_node": {
```
pregnor, line 72: **Note**: Base may have been a better term than regular, but whatever, it's not worth holding up over this, I will fix it if we get that far.

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
```diff
@@ -161,36 +251,92 @@ class Node(BaseModel):
...
+    """
+    if isinstance(value, (AgentNode, PartialAgentNode)):
+        return _AGENT_TAG
+    if isinstance(value, (RegularNode, PartialRegularNode)):
+        return _REGULAR_TAG
+    if not isinstance(value, dict):
+        return None
+    if "agent_record_uri" in value or "stable_agent_id" in value:
+        return _AGENT_TAG
+    return _REGULAR_TAG
```
pregnor, line 299: **Question**: Why didn't we build a discriminator from the `node.type` value? And instead of calling every node `customNode` we could have an agent type that could handle this discrimination and not need n+1 different types, only the agent extension with the required fields.

pregnor (commented):
Mostly looking good, couple questions.

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/event_v1.json
mihaialexandrescu: You're right. `base_node` would be better. I can take care of that later but I'd rather not further clutter this PR.

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
mihaialexandrescu: Right now the UI uses and "expects": `transportNode` or `customNode`.  
The current form of what we (temporarily) pre-load as "starting workflows" keeps the structure statically defined in the frontend.  
I tried to err/stay more within the bounds of what we have and work with that. That's the core explanation. It's not a "technical" one.

We could define enums for `node.type` and use that for the discriminator but that's some design that, I think, would have to be sync'ed across multiple components including frontend and be very intentional and designed.  
If we go down such a path, I would still do it in a different issue and PR because this one is already dilluted enough, IMO.

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
mihaialexandrescu: I don't follow. This is that portion of the schema. To me it looks like it is followed correctly (including at `description` level).

https://github.com/agntcy/coffeeAgntcy/blob/6a5cc196b3426e2dd313e62a396633f2081bed9f/coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/event_v1.json#L44-L52

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
mihaialexandrescu: I checked. I didn't find any other references to the simple `Data()` in the code.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_instance_store/store.py
mihaialexandrescu: I didn't notice.
Fixed in 02f8255d83d08af82458c7f26c16877803807038.

.agents/skills/jsonschema-to-pydantic-lungo/mapping-rules.md
mihaialexandrescu: If I understand the question correctly:

1. They are just examples - they don't even need to exist precisely that way in the jsonschema files
2. We can (and do) reference some portion of a markdown file from another markdown file but I can't find any evidence of being able to do that when targeting a JSON file

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
delthazor: Okay, I might have checked against the wrong json version, but if it is not present in the current, then it's ok.

delthazor (approved):
lgtm

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/event_v1.json
pregnor: Yeah, just ticket it and we will get to this later, post a ticket reference here and feel free to resolve the thread.

pregnor: I realized this is an AITA moment and did it: https://github.com/agntcy/coffeeAgntcy/issues/544

coffeeAGNTCY/coffee_agents/lungo/schema/types/event.py
pregnor: Makes sense, thanks for clarifying, created a refactor ticket for it for later: https://github.com/agntcy/coffeeAgntcy/issues/545

pregnor (approved):

coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas/event_v1.json
mihaialexandrescu: Not at all.  
Thank you for ticketing it!

Issue #450, opened 2026-03-16
[Feat] BLS Implement agentic workflow API instance/state subAPI endpoints


pregnor:
Phase 3, State API, SSE, and A2A middleware

Implement the instance/state endpoints of the agenric workflow API to return data of the instance state.
Reads from the workflow instance (in memory) state store only.

- Implement the instance/state endpoints from #447 :
    - `POST /agentic-workflows/{workflow_name}/`
    - `GET /agentic-workflows/{workflow_name}/instances/`
    - `GET /agentic-workflows{workflow_name}/instances/{workflow_instance_id}/`
- Use the store’s API from #447 / #448 . Response shape should be consistent with how “current state” is represented in the store and, where relevant, with the workflow instance state and event schema (#446) for consistency with SSE.
- Topology may evolve per workflow-instance (e.g. recruiter); this endpoint returns the current view for that workflow-instance.

PR #539, opened 2026-05-04
Implement sub-API middleware endpoints for workflow instance states

delthazor:
# Description

Implement proper endpoints workflow instance state reporting and managements that connects the API calls with the in-memory store.

Added e2e test that covers the intended call-flow through the system.

## Issue Link

Implements #450

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/instance_lifecycle.py
```diff
@@ -0,0 +1,105 @@
...
+"""Helpers for workflow instance HTTP lifecycle (seed ``event_v1``, read merged state)."""
+
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from uuid import uuid4
+
+from schema.types import Data, EventType, Workflow, WorkflowInstance
+
+_EVENT_V1_SCHEMA_VERSION = "1.0.0"
```
pregnor, line 13: **Question**: Is it worth tracking to the patch version?
We already modifying non-backward compatible stuff, we are around 1.2, 1.3 already, but we are only tracking major so far.

pregnor (dismissed):
LGTM

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/instance_lifecycle.py
delthazor: On the long run I think yes, but you are right about it being an edge case. Usually major version should suffice.

BTW We should've tracked patch versions, but until the system first fires up end-to-end, I think we can get away with keeping it as version 1.0.0. Once there are real consumers/listeners in the system, it would be wise to start proper bookkeeping on the minor/patch versions.

pregnor (dismissed):

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/instance_lifecycle.py
pregnor: Yeah, but minor and patch are no problems, because they are not supposed to break backward compatibility.

If we add a field or a new type that won't bother existing consumers.

Hence I thought it's enough to code the major version, hence the event_v1.json schema name.

delthazor: Yeah, I see your point, and you are probably right, I just would like to keep the possibility of checks like: `if(version > TARGET_PATCH): processExtraField()`
It might be unrealistic, to ever have something like that, but the option does not hurt :)

pregnor: That's fine, I just wanted to understand the logic behind, carry on.

pregnor (approved):

codyhartsook (approved):
lgtm

PR #543, opened 2026-05-04
feat(lungo): add Skill for OpenAPI spec to Python code and Pydantic types

mihaialexandrescu:
# Description

This PR aims to add a Skill specific to the Lungo subproject that can take the OpenAPI spec files and generate Pydantic v2 models from them for DTOs and a FastAPI endpoints router from the paths.

## Issue Link

Fixes #512 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

## Exception for the `instanceId` when used as a dict key

> **Known exception — flag in output.** `propertyNames` is **not enforced at the DTO layer**. Re-implementing the constraint here (for instance copying the `InstanceId` regex from `schema/types/event.py` into a `field_validator`) would duplicate logic that already lives in `schema/types/` and silently drift the next time the canonical type changes. The skill deliberately relies on the value type's own validation: a `WorkflowInstance` already validates its `id: InstanceId` field, so any caller that constructs a value with a mismatched key gets a runtime check on the nested id (and the cross-check that key equals nested id is, today, the responsibility of the value type itself — see `Workflow.instances` in `schema/types/event.py`). When generating or validating a DTO with a constrained `propertyNames`, mention this exception in the output and recommend revisiting it once `schema/types/` exposes its key patterns as importable constants (which would let the skill enforce the constraint without duplication).

> Why never `dict[InstanceId, V]` (or any other Pydantic `RootModel` subclass as the key):
> 
> - `RootModel` instances are unhashable by default (`__hash__ = None`); `model_validate(...)` on the map raises `TypeError: unhashable type` before it can populate anything.
> - Setting `model_config = ConfigDict(frozen=True)` on the key model fixes hashability but `model_dump_json` then serializes keys as the model's Python repr (e.g. `"root='instance://...'"`), which violates the OpenAPI contract that JSON object keys are bare strings.

I currently chose the option that yields the same code we have today. IMO we might want to deal with this in a dedicated fashion later on because this could be a recurring thing.  
To that end:
  * The Skill that generates the `event.py` Pydantic v2 models from the JSONSchema could take care to also do the improved `repr` method override. That's a good idea in general, I think, for those "...Id" types to have them serialize in a more intuitive fashion to our data external models. 
  * The `frozen=True` setting would be needed to make those "...Id" types hashable but could impact our 2-step ID generation/population... this may look "easy" but I don't know that it is obvious that it won't bite us later on. I also don't know if we should maybe only apply this to models that are used as dict keys..? of which we may only have 1.
   Right now we duplicate the field validation even in `event.py` for `instanceId`.

Another thing to consider/look into:  if we give up the `RootModel` based pattern for the "...Id" types and we replace it with an alias for the otherwise identical `Annotated[str, ...]` there is a chance that both issues are addressed or improved. I did not have time to try it out. I wanted to get the review on the other "Skill"-centric notions first/quicker.
   
The first commit in the PR has a version where such logic gets replicated into a `field_validator` in the `dtos.py` file if you want to inspect that, too.

## Other caveats

**Framework DI params not in the spec**: several handlers take `request: Request` and `delete` takes `background_tasks: BackgroundTasks`. The skill's signature rules only cover path/query/body params (found in the spec), so these need preserving.

**Factored helper modules**: `main` split logic into `auth.py`, `instance_lifecycle.py`, `patterns.py`, `use_cases.py`, `workflow_documentation.py`, `workflows.py`, `server.py`. The skill's model is a single `router.py` + `dtos.py` per tag. The skill preserves handler bodies and module-level helpers, but it has no concept of these sibling files — fine for `dtos.py`/`router.py` equivalence, but a regen wouldn't recreate that structure.


delthazor (dismissed):
lgtm

codyhartsook (dismissed):
lgtm

pregnor (dismissed):

delthazor (approved):
lgtm

pregnor (approved):
LGTM

PR #548, opened 2026-05-05
feat(lungo): propagate workflow context across supervisors (#453)

### LINKED ISSUE #453: [Feat] BLS Emit workflow-instance state progress from LangGraph/agents via A2A to state middleware

codyhartsook:
Refs: [#453](https://github.com/agntcy/coffeeAgntcy/issues/453)

## Summary

This PR adds workflow-context propagation so downstream emitters can read `workflow_name` and `workflow_instance_id` without threading identity through every signature. This is the foundation for the A2A workflow-event emission work in #528.

### What changed

- Adds `common/workflow_context_prop.py` with attach / detach / scope / read helpers backed by OpenTelemetry baggage.
  - `contextvars`-isolated per asyncio Task, so concurrent requests don't bleed across each other.
  - W3C-standard header so the same identity flows across A2A hops.
- Wires the helpers into all three supervisor entrypoints:
  - **Auction** (`ExchangeGraph.serve`, `ExchangeGraph.streaming_serve`)
  - **Logistics** (`LogisticGraph.serve`, `LogisticGraph.streaming_serve`)
  - **Recruiter** (`call_agent`, `stream_agent`)
- Resolution priority at each entrypoint: upstream OTel baggage → caller `workflow_instance_id` kwarg → minted `instance://<uuid4>` fallback.
- Async generators (`streaming_serve`, `stream_agent`) use manual `attach_workflow_context` / `detach_workflow_context` in `try / finally` rather than `with workflow_context_scope(...)` to avoid OTel "Failed to detach context" warnings on `aclose()` / early break (async-generator `__exit__` can run on a different Task than `__enter__`).
- Pins workflow names per supervisor as module-level constants:
  - Auction serve → `Publish Subscribe Coffee Farm Network`
  - Auction stream → `Publish Subscribe Streaming Coffee Auction Network`
  - Logistics → `Secure Group Communication Logistics Network`
  - Recruiter → `On-demand Discovery`
- Adds drift-guard tests under `tests/unit/{auction,logistics,recruiter}/test_*_workflow_names.py` that load `api/agentic_workflows/starting_workflows.json` and assert each constant exists in the catalog.

## Testing

- `uv run pytest tests/unit/auction/test_auction_workflow_names.py tests/unit/logistics/test_logistics_workflow_names.py tests/unit/recruiter/test_recruiter_workflow_names.py -v`

## Follow-ups (tracked separately)

- Wire `auction/main.py`, `logistics/main.py`, `recruiter/main.py` to attach context from inbound HTTP requests so per-supervisor fallbacks become unreachable.
- Implement `POST /agentic-workflows/{name}/` to mint canonical instance ids upstream (currently 501).

Closes #453


coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
```diff
@@ -35,6 +41,11 @@
 _SUPERVISOR_FARM_KEY = "supervisor_farm"
 _SUPERVISOR_OPERATION_KEY = "supervisor_operation"
 
+# Must match entries in api/agentic_workflows/starting_workflows.json.
+# Drift is guarded by tests/unit/auction/test_workflow_names.py.
```
delthazor: I think the real test file name is test_auction_workflow_names.py
I see the intent, but I generally advise against file names in comments, since they can easily change.
Consider something like "Drift is guarded by corresponding auction unit tests" or similar.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/logistics/graph/graph.py
```diff
@@ -17,9 +17,19 @@
...
+    attach_workflow_context,
+    detach_workflow_context,
+    read_workflow_context,
+    workflow_context_scope,
+)
 
 logger = logging.getLogger("lungo.logistics.supervisor.graph")
 
+# Must match starting_workflows.json. Drift guarded by
+# tests/unit/logistics/test_workflow_names.py.
```
delthazor: Same here with logistics. See my comment about auction test

coffeeAGNTCY/coffee_agents/lungo/common/workflow_context_prop.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Call from a supervisor request handler when the caller provides a
+    workflow_instance_id from the Agentic Workflows API. Pair with
+    detach_workflow_context in a finally block, or use
+    workflow_context_scope to handle attach/detach automatically.
+
+    Returns None when neither value is provided.
+    """
+    if not workflow_instance_id and not workflow_name:
+        return None
```
delthazor: I'm not sure if returning None instead of a valid token, is the right approach. It doesn't really make sense to call this with both params set to None, so maybe an error/exception would be a more appropriate response to an invalid input.
I might be mistaken about this case though, so correct me if I'm missing some context here that validates the returned None.

delthazor (commented):
The tests added seems to have some overlap, or at least similar setup/boilerplate. If it is feasible, consider factoring them out into. test helper, to avoid duplications.

pregnor (commented):
Aside from Peter's comments - which I agree with - LGTM.

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/graph.py
codyhartsook: updated comment

coffeeAGNTCY/coffee_agents/lungo/common/workflow_context_prop.py
codyhartsook: replaced return None with raise

pregnor (approved):

mihaialexandrescu (approved):

PR #571, opened 2026-05-18
Fix backend leak by adding DELETE workflow endpoint

delthazor:
# Description

Currently, new workflow instances are created every time the user navigates to a menu item, but they are never cleaned. This change aims to make memory management more safe, by adding and calling a proper DELETE endpoint, and protecting it with a secret, so not anyone can call it just the authenticated ui endpoints can use it.

## Issue Link
No corresponding issue.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


pregnor:
Ugh, you put me in a very difficult spot.

Normally I would make sure this is included in the release, but noticing this the release day +7 and the sheer size and decisions made in this PR I cannot promise that this can go in today due to the time pressure, sorry.

We can do a hotfix release after, but I really don't feel like I can risk this without proper focused, non-rushed review.

delthazor:
> Ugh, you put me in a very difficult spot.
> 
> Normally I would make sure this is included in the release, but noticing this the release day +7 and the sheer size and decisions made in this PR I cannot promise that this can go in today due to the time pressure, sorry.
> 
> We can do a hotfix release after, but I really don't feel like I can risk this without proper focused, non-rushed review.

It is perfectly fine if only merged later, it takes a while for the instances to accumulate, and it seems the CI ha problems with the tests, I can not reproduce locally yet, so it still needs some fine-tuning.

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/paths/agentic-workflows.yaml
```diff
@@ -200,6 +200,18 @@ paths:
...
+      operationId: deleteWorkflowInstance
+      summary: Delete workflow instance
+      description: >
+        Removes the instance from the in-memory store. First delete returns 204;
+        a second delete for the same id returns 404.
+      responses:
+        "204":
+          description: Instance removed
+        "404":
+          description: Workflow or instance not found
```
pregnor: **Question**: Shouldn't we be rather using 202 ACCEPTED for existing resources and 204 NO CONTENT on non-existent resources?

My reasoning for it is to have idempotent functionality - something I'm gunning for in most ops - so the user doesn't have to do 2 calls - check if existing and then call delete - if they want a non-erroneous delete without prior knowledge of existence - such as due to distributed processing nature.

Also regarding codes, for not-yet-done deletion I prefer 202 because it signals that the server ACKed the request and eventually delete will happen, maybe started and ongoing, maybe pending.
And when called on a non-existent resource we don't know whether it never existed or just has been deleted 5s ago or a year ago, it's safer to say we processed the request, nothing to do, carry on.
In "scripting" (batch-processing) context this can significantly help the caller's work because it streamlines ID based deletion looping while also reflects on resource existence in the success return code.

coffeeAGNTCY/coffee_agents/lungo/docs/publish_subscribe_workflow_walkthrough.md
pregnor, line 1: **Question**: What is this file, do we need it?

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/event_sink.py
```diff
@@ -38,13 +38,19 @@ class WorkflowAPIEventSink(EventSink):
...
-    def __init__(self, base_url: str | None = None) -> None:
+    def __init__(self, base_url: str | None = None, api_key: str | None = None) -> None:
         if base_url is None:
             from config.config import WORKFLOW_API_URL
 
             self._base_url = WORKFLOW_API_URL
         else:
             self._base_url = base_url
+        if api_key is None:
+            from config.config import WORKFLOW_API_KEY
```
pregnor: **Required**: We **MUST** make the agents stop generating function level imports like this.

coffeeAGNTCY/coffee_agents/lungo/deployment/helm/agentic-workflows-api/values.yaml
```diff
@@ -18,6 +18,12 @@ service:
 
 config:
   corsAllowedOrigins: ""
+  # Set for local/dev chart installs only; production should use an external Secret.
+  workflowApiKey: ""
```
pregnor: **Question**: Could we add a dummy value here for default? What are the drawbacks?

Asking because if we implement the auto-deployment this would break things as the secret value would be missing initially before defining it manually on the deployment values and a default value would avoid that.

I know from security perspective it's more likely people forget to change the default value if we provide one, but I'm not sure I want to factor that in.

(gosh, we also need to refactor the config interface at some point to a more sane structure)

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
```diff
@@ -354,6 +359,44 @@ async def get_workflow_instance_state(
...
+                status_code=404, detail=f"Workflow not found: {workflow_name}"
+            )
+        canonical_id = _canonical_instance_uri(workflow_instance_id)
+        store = _workflow_instance_store(request)
+        deleted = await asyncio.to_thread(
+            delete_workflow_instance_from_store,
+            store,
+            workflow_name,
+            canonical_id,
+        )
```
pregnor: **Question**: Shouldn't we just fire this async without waiting?
We shouldn't hold the caller up during the deletion period - especially so, because most stuff should be moved to a soft-delete approach later on in general.

coffeeAGNTCY/coffee_agents/lungo/.env.example
```diff
@@ -91,6 +91,13 @@ HOT_RELOAD_MODE=true
 # Lungo UI (`docker compose --profile frontend up --build` from lungo/): **`VITE_*` live only in
 # `frontend/.env`** — copy from `frontend/.env.example`. Do not duplicate here.
 
+# === Agentic Workflows API (port 9105) ===
+# Shared secret for POST/GET/DELETE on the workflow instance API and internal POST .../events/.
+# Set the same value as VITE_AGENTIC_WORKFLOWS_API_KEY in frontend/.env for the UI build.
+# WORKFLOW_API_KEY=
+# Base URL for agents that emit workflow events (compose: http://agentic-workflows-api:9105).
+# WORKFLOW_API_URL=http://agentic-workflows-api:9105
```
pregnor: **Required**: We **MUST** uncomment required env vars in the example to be easy to comprehend what to set. (at least the `WORKFLOW_API_KEY`).

Same question here as in the charts, shouldn't we set a placeholder value to the key, named appropriately?

coffeeAGNTCY/coffee_agents/lungo/tests/unit/agentic_workflows/api/test_workflow_instance_state_endpoints.py
```diff
@@ -267,3 +276,108 @@ def test_multiple_instantiations_list_two_instances(client: TestClient) -> None:
     assert g2.status_code == 200
     assert g1.json()["id"] == wid1
     assert g2.json()["id"] == wid2
+
+
+def test_delete_instance_returns_204_then_404(client: TestClient) -> None:
```
pregnor: **Recommended**: We **SHOULD** stick to the table testing in this case IMO, but feel free to refute.

pregnor (changes requested):

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/paths/agentic-workflows.yaml
delthazor: That sounds reasonable

coffeeAGNTCY/coffee_agents/lungo/docs/publish_subscribe_workflow_walkthrough.md
delthazor: Nope, that should not have been added, I will remove that one.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/event_sink.py
delthazor: Agreed

coffeeAGNTCY/coffee_agents/lungo/deployment/helm/agentic-workflows-api/values.yaml
delthazor: The only drawback I know of, is the same you mentioned: forgetting to change it impose a security risk. If you think this is a risk that can be taken, I can add a default value.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
delthazor: I agree, that would be better to call async

coffeeAGNTCY/coffee_agents/lungo/tests/unit/agentic_workflows/api/test_workflow_instance_state_endpoints.py
delthazor: I'll see to it.

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/paths/agentic-workflows.yaml
delthazor: Updated as requested

coffeeAGNTCY/coffee_agents/lungo/docs/publish_subscribe_workflow_walkthrough.md
delthazor: Removed

coffeeAGNTCY/coffee_agents/lungo/common/a2a_event_middleware/event_sink.py
delthazor: I fixed that, and I'm still looking into ways to configure agents to stop generating such imports.

coffeeAGNTCY/coffee_agents/lungo/deployment/helm/agentic-workflows-api/values.yaml
delthazor: Added a default value

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/router.py
delthazor: Made async: call and go on, no wait.

coffeeAGNTCY/coffee_agents/lungo/.env.example
delthazor: Uncommented required rows, added the default value mentioned previously

pregnor (approved):

mihaialexandrescu (approved):

PR #579, opened 2026-05-19
fix(transport): default SLIM timeout to 20s, make configurable via env

### LINKED ISSUE #519: [Bug]: SLIM request() default timeout (6s) is too short for LLM-backed farm agent responses

mairp:
# Description

Patch SLIMTransport.request default timeout from the SDK's 6s to 20s in all agent servers and the shared a2a transport config. Also surface real backend errors in the frontend by parsing data.detail.

## Why 20s?

The SLIM SDK ships with a **6-second default timeout** for `SLIMTransport.request`. This is too aggressive for agent-to-agent calls where the downstream agent invokes an LLM — typical LLM inference takes **5-20 seconds** depending on model, prompt complexity, and provider load.

| Value | Trade-off |
|-------|-----------|
| 6s (SDK default) | Fails on nearly every real LLM-backed agent call |
| 10s | Still tight — clips longer prompts and multi-tool chains |
| **20s** | **Covers the P95 of single-hop LLM calls while surfacing genuine failures in under half a minute** |
| 60s+ | Too generous — users wait a full minute before seeing a real error |

**20s is the sweet spot**: it eliminates the false-positive timeouts reported in #519 without masking real connectivity failures. The value is also **fully configurable** via `SLIM_TRANSPORT_REQUEST_TIMEOUT` env var, so deployments with different latency profiles can tune it without code changes.

## What changed

- Default timeout in `corto/config/config.py` and `lungo/config/config.py`: **6s → 20s**
- `corto/.env` previously hardcoded `6` (defeating the fix) — updated to `20`
- `.env.example` files in both projects updated to reflect the new default
- Timeout is monkey-patched via `SLIMTransport.request.__defaults__` after all imports in every server entry point
- Frontend: real backend errors surfaced by parsing `data.detail`

## CI note

The integration test failures (`test_sommelier`, `test_auction_*`, `test_logistics_*`) are **not caused by this PR**. They fail because `LLM_MODEL` is empty in the CI environment — the `litellm.BadRequestError: LLM Provider NOT provided. You passed model=` error confirms the LLM secrets are not available in the PR workflow context. These same tests pass on `main` where secrets are injected.

## Issue Link

Fixes #519 and #518

## Type of Change

- [X] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


coffeeAGNTCY/coffee_agents/lungo/agents/logistics/accountant/server.py
```diff
@@ -19,6 +19,10 @@
     SlimTransportConfig,
 )
 from agntcy_app_sdk.semantic.a2a.client.factory import A2AClientFactory
+from agntcy_app_sdk.transport.slim.transport import SLIMTransport
+
+SLIMTransport.request.__defaults__ = (60,)
```
delthazor: 1.
I don't think subsequent imports need this to be defined, so putting it in the middle, instead of after the import block is not recommended.

2.
Also, if it repeats as such, it should not be copied as a magic number. There should be a common constant for it. In addition, if it is already touched, it should be configurable, so the constant should come from config / .env

3.
Why 60? It seems unnecessarily long for the given circumstances. Ideally, such messages should not take longer than 8-10 secs to round-trip, and for most part, exp-backoff is already in place. So even if it reaches a timeout due to some network congestion, it should be successful for one of the retries, when traffic becomes lower in a few secs. I think even 30s would be overkill, and would make scenarios where there is a real problem behind, take very long to notice due to timeout. If you insist on such a high number, please attach some reasoning, preferably with measurements.

delthazor (changes requested):

coffeeAGNTCY/coffee_agents/lungo/agents/logistics/accountant/server.py
mairp: Thanks for the feedback! Addressed all three points in the latest push:

1. **Import placement**: Moved the `SLIMTransport.request.__defaults__` override after all imports — no longer sitting in the middle of the import block.

2. **Configurable, no magic numbers**: The timeout is now driven by a `SLIM_TRANSPORT_REQUEST_TIMEOUT` env var read from `config/.env`. Every server entry point reads from `config.py`, eliminating the duplicated hardcoded value.

3. **Timeout value**: Agreed — 60s was too high. When `SLIM_TRANSPORT_REQUEST_TIMEOUT` is not set, the SDK default (6s) applies unchanged — no monkey-patch runs at all. The `.env.example` documents 6s as the default, leaving it up to the operator to tune as needed for their deployment.

delthazor (approved):
I approved it before, thought I realized, the defaults are not changed with the change.

coffeeAGNTCY/coffee_agents/lungo/config/config.py
```diff
@@ -34,6 +34,9 @@
 OAUTH2_BASE_URL= os.getenv("OAUTH2_BASE_URL", "")
 OAUTH2_APPKEY= os.getenv("OAUTH2_APPKEY", "")
 
+_slim_timeout_raw = os.getenv("SLIM_TRANSPORT_REQUEST_TIMEOUT")
+SLIM_TRANSPORT_REQUEST_TIMEOUT = int(_slim_timeout_raw) if _slim_timeout_raw is not None else None
+
```
delthazor: Actually this way, the original issue is not addressed, the 6s default timeout is small. It would be better to have like 10s (or max 15s).

coffeeAGNTCY/coffee_agents/lungo/.env.example
```diff
@@ -80,6 +80,8 @@ NATS_SERVER="nats:4222"
 
 SLIM_SHARED_SECRET="slim-shared-secret-REPLACE_WITH_RANDOM_32PLUS_CHARS"
 
+# SLIM transport request timeout in seconds (overrides SDK default when set)
+# SLIM_TRANSPORT_REQUEST_TIMEOUT=6
```
delthazor: Even if it is commented out, the placeholder value should reflect the used default. Once you addressed my comment above, the same value should be inserted in the env examples.

delthazor (changes requested):

coffeeAGNTCY/coffee_agents/lungo/config/config.py
mairp: Good point — defaulting to `None` left the original 6s problem in place for anyone not setting the env var.

Changed the default to **10s**: it's ~1.6× the SDK default which gives enough headroom for transient network congestion or slow SLIM round-trips, while still surfacing real connectivity issues quickly (a genuine failure won't hide behind a 10s wait when exp-backoff retries are already in play). The value remains overridable via `SLIM_TRANSPORT_REQUEST_TIMEOUT`.

Also simplified the monkey-patch sites — since the timeout is now always an `int`, the `if is not None` guard was dead code.

coffeeAGNTCY/coffee_agents/lungo/.env.example
mairp: Updated — both `.env.example` files now show `SLIM_TRANSPORT_REQUEST_TIMEOUT=10`, matching the new default.

coffeeAGNTCY/coffee_agents/lungo/agents/logistics/shipper/server.py
```diff
@@ -31,6 +33,9 @@
 from starlette.routing import Route
 from uvicorn import Config, Server
 
+if SLIM_TRANSPORT_REQUEST_TIMEOUT is not None:
+    SLIMTransport.request.__defaults__ = (SLIM_TRANSPORT_REQUEST_TIMEOUT,)
+
```
delthazor: I think the check is no longer needed. After the default value has been set, I think there is no way for the timeout config param to be None.

coffeeAGNTCY/coffee_agents/lungo/config/config.py
delthazor: I accept this as resolved.

coffeeAGNTCY/coffee_agents/lungo/.env.example
delthazor: Accepted as resolved.

coffeeAGNTCY/coffee_agents/lungo/agents/logistics/shipper/server.py
mairp: Good catch — since `SLIM_TRANSPORT_REQUEST_TIMEOUT` now always has a default value, the `None` guard was indeed dead code. Removed it from `farm/server.py`, `helpdesk/server.py`, and `shipper/server.py` so they match the unconditional pattern already used in the other entry points.

delthazor (approved):
seems fine now

pregnor (approved):

pregnor:
@delthazor will look into the integration test failures, how to fix the external contributor lack of secrets situation in a safe way, thanks for your patiences regarding this, otherwise the PR is mergable from our side.

@codyhartsook if you want to take a look at the PR in the meantime, feel free.

pregnor:
We will track the CI unblocking in https://github.com/agntcy/coffeeAgntcy/issues/595

mihaialexandrescu:
Thank you for your contribution!

Our repository has moved quite a bit in the mean time before we got to review this PR. I didn't want to make you go through the hassle of bringing the PR up to date so I created a different PR (https://github.com/agntcy/coffeeAgntcy/pull/768) with your contribution, solving some merge conflicts and some style differences we now watch for and some additional frontend error message handling.
As I mention in that PR's description, the credit goes to you.

mairp:
Appreciate you taking this upstream, @mihaialexandrescu . Given how much drift there'd been since this was opened, folding it into a clean PR made way more sense than asking for a rebase. Thanks for keeping the attribution straight and for the extra work on the frontend error handling in #768. Good call, and nice work getting it shipped.

PR #581, opened 2026-05-22
refactor: use uv docker base image for agents

arpad-csepi:
# Description

This PR change-set includes improvements across the dockerfiles, mainly focusing on the optimization of the agents building. This is achieved by using `uv` docker image as a base image and remove some steps from docker image building.

The improvements measured on a M1 Pro Macbook Pro was around 30s (2min instead of 2.5 min) with `farms,observality,frontend` compose profiles.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [X] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
@@ -16,12 +16,6 @@ ARG BUILD_SOURCE=local
 ARG IMAGE_NAME=unknown
 ARG IMAGE_TAG=unknown
 
-# Install system dependencies, including curl, ping, and wget
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends git curl wget iputils-ping && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
 WORKDIR /app
 COPY coffeeAGNTCY/coffee_agents/corto/pyproject.toml coffeeAGNTCY/coffee_agents/corto/uv.lock ./
 RUN uv sync --locked --no-editable
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
@@ -17,12 +17,21 @@ ARG IMAGE_NAME=unknown
 ARG IMAGE_TAG=unknown
 
 WORKDIR /app
-COPY coffeeAGNTCY/coffee_agents/corto/pyproject.toml coffeeAGNTCY/coffee_agents/corto/uv.lock ./
 
+# Silences warnings about not being able to link files since the cache and sync target are on separate file systems
 ENV UV_LINK_MODE=copy
-RUN --mount=type=cache,target=/root/.cache/uv  uv sync --locked --no-editable
+
+# Install dependencies
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=coffeeAGNTCY/coffee_agents/corto/uv.lock,target=uv.lock \
+    --mount=type=bind,source=coffeeAGNTCY/coffee_agents/corto/pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project
 
 COPY coffeeAGNTCY/coffee_agents/corto/ .
+
+# Sync the project
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked
 
 ## Generate about.properties file with build metadata
 RUN printf "app.name=corto-exchange\n"                > about.properties && \
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
@@ -18,7 +18,9 @@ ARG IMAGE_TAG=unknown
 
 WORKDIR /app
 COPY coffeeAGNTCY/coffee_agents/corto/pyproject.toml coffeeAGNTCY/coffee_agents/corto/uv.lock ./
-RUN uv sync --locked --no-editable
+
+ENV UV_LINK_MODE=copy
+RUN --mount=type=cache,target=/root/.cache/uv  uv sync --locked --no-editable
 
 COPY coffeeAGNTCY/coffee_agents/corto/ .
 
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
@@ -1,6 +1,6 @@
 # `coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange`
 # SLIM v0.7+ requires to use debian trixie distribution
-FROM python:3.13.11-slim-trixie
+FROM astral/uv:0.11.16-python3.13-trixie-slim@sha256:d57fc364ed714127c162a01b316bfb7b9a3fd09d97c665781a5607acaa35d254
 
 # Build arguments for metadata
 ARG BUILD_VERSION=unknown
@@ -21,11 +21,6 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends git curl wget iputils-ping && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
-
-# Install uv and create virtual environment
-RUN pip install --no-cache-dir uv && \
-    uv venv -q /venv
-ENV PATH="/venv/bin:$PATH"
 
 WORKDIR /app
 COPY coffeeAGNTCY/coffee_agents/corto/pyproject.toml coffeeAGNTCY/coffee_agents/corto/uv.lock ./
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
@@ -1,16 +1,11 @@
 # SLIM v0.7+ requires to use debian trixie distribution
-FROM python:3.13.11-slim-trixie
+FROM astral/uv:0.11.16-python3.13-trixie-slim@sha256:d57fc364ed714127c162a01b316bfb7b9a3fd09d97c665781a5607acaa35d254
 
 # Install system dependencies, including ping, curl, and wget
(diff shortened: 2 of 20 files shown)


coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
```diff
@@ -1,6 +1,6 @@
 # `coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange`
 # SLIM v0.7+ requires to use debian trixie distribution
-FROM python:3.13.11-slim-trixie
+FROM astral/uv:0.11.14-python3.12-trixie-slim@sha256:13b5883729ec534af5863facf5c40ac0869f2c124ad5620c9ed36fe7c03fa87d
```
delthazor: Why downgrade from python 3.13 to 3.12?

coffeeAGNTCY/coffee_agents/lungo/docker/Dockerfile.ui
```diff
@@ -2,25 +2,21 @@ FROM node:20-alpine
...
 # Copy application files, including assets
 COPY coffeeAGNTCY/coffee_agents/lungo/frontend/ .
 
-# This build step is simply for checking the validity of the build:
-# running type checks, linting, and making sure the build is functional.
-RUN ["npm", "install", "--verbose"]
-RUN ["npm", "run", "build", "--verbose"]
-RUN ["npm", "cache", "clean", "--force"]
+# Build the frontend
+RUN npm install --verbose
```
delthazor, line 15: Why `npm install` after `npm ci` ? The ci version should be sufficient, especially since the other one can drift away from package lock.

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
arpad-csepi: My bad, fixed.

coffeeAGNTCY/coffee_agents/lungo/docker/Dockerfile.ui
arpad-csepi: The npm ci used with only with the `package.json` to install the dependencies from it as it defined there but I run it before copying the source files to the container. This way the docker build cache can hold the cached packages longer as a source code change does not invalidate it.

I can use only `npm ci` instead this approach still if it's preferred.

delthazor: I think that would be the better one, but it is not a strict policy for the project. I would prefer it, but you are free to discard.

arpad-csepi: I think it is more useful to keep the downloaded python packages in docker build cache based on the `package.json` and `package-lock.json` so when only the python source code is changed the build would be faster. Because of this I prefer to keep this solution.

coffeeAGNTCY/coffee_agents/lungo/docker/Dockerfile.ui
```diff
@@ -2,25 +2,21 @@ FROM node:20-alpine
...
 
 # Expose port 3000
 EXPOSE 3000
 
-# Copy the script to build and serve the UI
-COPY coffeeAGNTCY/coffee_agents/lungo/docker/ui-build-and-serve.sh ./ui-build-and-serve.sh
 # Start the application
-CMD ["sh", "./ui-build-and-serve.sh"]
\ No newline at end of file
+CMD ["npx", "serve", "-s", "dist", "-l", "3000"]
```
pregnor, line 22: **Question**: I know I'm overstaying my welcome, but I checked and this is the only place lungo was using the ui-build-and-serve.sh, could we remove it as well as part of this PR?

Also we might want to do a similar change on the Corto Dockerfile.ui which is already using cache, but uses an identical shell script, I would like to incorporate the instructions into the dockerfile from there as well.

My text search did not yield any other references to such a script, so we should be safe to remove those.

pregnor (dismissed):

coffeeAGNTCY/coffee_agents/lungo/docker/Dockerfile.ui
arpad-csepi: Done, please check it.

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
```diff
@@ -16,23 +16,23 @@ ARG BUILD_SOURCE=local
...
-    uv venv -q /venv
-ENV PATH="/venv/bin:$PATH"
+# Silences warnings about not being able to link files since the cache and sync target are on separate file systems
+ENV UV_LINK_MODE=copy
 
-WORKDIR /app
-COPY coffeeAGNTCY/coffee_agents/corto/pyproject.toml coffeeAGNTCY/coffee_agents/corto/uv.lock ./
-RUN uv sync --locked --no-editable
+# Install dependencies
+RUN --mount=type=cache,target=/root/.cache/uv \
```
mihaialexandrescu, line 25: This is **OPTIONAL** and very much up for debate:

We don't do rootless containers and we always run as `root` inside the containers so this approach holds.  
However, do you guys reckon it would be a bit better to be explicit about where the `uv` cache is expected to be found ? The options would be an `ENV UV_CACHE_DIR` statement or using the CLI flag on every `uv sync` (less preferable, in my opinion).

---

AI-generated context:

### How uv decides where its cache lives

`uv` resolves its cache directory in this order of precedence (highest first):
  * `--cache-dir <path>` CLI flag — if passed, this wins.
  * `UV_CACHE_DIR` environment variable — if set, this is used.
  * Platform default, derived from XDG / OS conventions:
    * Linux: `$XDG_CACHE_HOME/uv`, falling back to $HOME/.cache/uv when XDG_CACHE_HOME is unset.
    * macOS: `~/Library/Caches/uv`.
    * Windows: `%LOCALAPPDATA%\uv\cache`.

For example, In the Recruiter Dockerfile, none of `--cache-dir`, `UV_CACHE_DIR`, or `XDG_CACHE_HOME` is set. The image runs as `root`, so `$HOME = /root`, and the Linux default becomes: `$HOME/.cache/uv  →  /root/.cache/uv`

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
```diff
@@ -1,20 +1,21 @@
...
-    uv venv -q /venv
-ENV PATH="/venv/bin:$PATH"
+# Silences warnings about not being able to link files since the cache and sync target are on separate file systems
+ENV UV_LINK_MODE=copy
 
-WORKDIR /app
-COPY coffeeAGNTCY/coffee_agents/corto/pyproject.toml coffeeAGNTCY/coffee_agents/corto/uv.lock ./
-RUN uv sync --locked --no-editable
+# Install dependencies
+RUN --mount=type=cache,target=/root/.cache/uv \
```
mihaialexandrescu, line 10: This is **OPTIONAL** and very much up for debate:

Seeing as the dependencies may not always be identical between our subprojects (corto, lungo, recruiter), would it be better to `id` the `uv` cache backing folder per project (see explanation and syntax below) ?  

This is optional because these days we mainly work on `lungo` so it doesn't matter so much if `corto` has different dependencies and the dependencies between `recruiter` and `lungo` are very likely to be in-sync.

---

AI-generated context:

### How BuildKit picks which host-side cache to attach

The mapping is controlled by the mount's `id`:
  * If you write `--mount=type=cache,target=/root/.cache/uv`, BuildKit uses the target value as the implicit `id`. So every `RUN` in every Dockerfile on that builder that says `target=/root/.cache/uv` ends up sharing the same host-side cache directory.
  * If you want isolation (say, a separate `uv` cache per project), use an explicit `id`: 
    * `RUN --mount=type=cache,id=recruiter-uv,target=/root/.cache/uv ...`
    * Now the host-side store keyed by `recruiter-uv` is what gets mounted at `/root/.cache/uv` inside the build step. 

Two different Dockerfiles using the same `id` will share; using different ids will not.

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
delthazor: I would say, the env var is preferred for cleanliness.

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
delthazor: I see the problem here, and I would recommend using explicit id for separation, but as far as I know, Corto is planned to be worked into Lungo in the long term plans, so they won't be separate projects (in the somewhat far future), which may make this pointless later. I don't know if this worth to be handled now with this in mind.

delthazor (approved):
lgtm

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
pregnor: I would do this in a separate PR, but IMO the env var implementation makes sense.

@mihaialexandrescu would you like to pick that up and do a PR for that? After Árpád you have most of the context and I don't want to task him more as this was already a free-time-contribution from him.

pregnor (approved):

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
pregnor: Good catch, this could be part of the separate PR I have mentioned above if you agree.

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
mihaialexandrescu: Sure thing.

Issue: https://github.com/agntcy/coffeeAgntcy/issues/593

coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
mihaialexandrescu: Sounds good. Will do.

Issue: https://github.com/agntcy/coffeeAgntcy/issues/593

mihaialexandrescu (approved):
good ideas!

Issue #569, opened 2026-05-18
[Bug]: Replace temporary static mappings with proper dynamic agent data


delthazor:
### Bug Description

While making the drawing of the graphs dynamic (based on backend catalog) instead of the previous static graph definitions, some node/graph data needed to be statically mapped back, as it was missing from the backend data. This needs to be replaced with proper backend-driven data management.
The corresponding PR where the change was introduced and discussed is here: https://github.com/agntcy/coffeeAgntcy/pull/551

### Expected Behavior

No statically mapped data for agents especially not form the UI side.

### Affected Version

v1.0.0

### Steps to Reproduce

no steps, just observe the corresponding code

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

Issue #568, opened 2026-05-18
Fix FE hardcoded workflow map


pregnor:
https://github.com/agntcy/coffeeAgntcy/pull/551#discussion_r3234602634

<img width="2560" height="1409" alt="Image" src="https://github.com/user-attachments/assets/d58b661f-fd24-4677-8707-c0e6b9a044cb" />

Replace it with the dynamic backend endpoint calling pls.

PR #586, opened 2026-05-26
Derive directory agent url dynamically

delthazor:
# Description

Currently a static map is used to connect the new agent graph nodes received from the backend, to the directory urls previously defined for them, when the graphs were statically defined. These urls can be derived from already received workflow summary, so static mapping is removed.

## Issue Link
Partially corresponds to #569   and  #568 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


pregnor:
@misi-bp look 🙂

pregnor (dismissed):
LGTM, but wait for @misi-bp 's review as well please.

coffeeAGNTCY/coffee_agents/lungo/tests/integration/conftest.py
```diff
@@ -380,10 +379,13 @@ def test_things(agents_up): ...
         print(f"--- {name} ready (logs: {runner.log_path}) ---")
         runners.append(runner)
 
+    if runners and AGENT_POST_READY_SETTLE_SECS > 0:
+        time.sleep(AGENT_POST_READY_SETTLE_SECS)
+
```
pregnor: **Recommended**: Maybe we **SHOULD** add an explanation on "why we need this" for later, because this is not intuitive.

pregnor (dismissed):

coffeeAGNTCY/coffee_agents/lungo/tests/integration/conftest.py
delthazor: Changed

pregnor (dismissed):

mihaialexandrescu (dismissed):

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
```diff
@@ -170,6 +143,28 @@ export function useWorkflowGraphFromAgenticApi({
...
+  const refetchAndApplyTopologyRef = useRef<
+    (seq: number, attempt: number) => Promise<void>
+  >(async () => {})
+  const scheduleTopologyRefetchRef = useRef<() => void>(() => {})
+
+  const agenticSessionRefs = useRef<WorkflowGraphAgenticSessionRefs | null>(
+    null,
+  )
+  if (agenticSessionRefs.current === null) {
+    agenticSessionRefs.current = {
```
misi-bp: If we use an umbrella-object in a ref to hold variables, why do we need all the 'subrefs' ?
I'm not sure we need both agenticSessionRefs and the 'subrefs', though it should be tested
(useRef is used IN components and provides the opportunity to reach variables/values with the same identity even when component is refreshed.)

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
```diff
@@ -251,7 +244,6 @@ export function useWorkflowGraphFromAgenticApi({
     [attachHandlers, setNodes, setEdges, restoreEdgeAnimation],
   )
 
-  const applyInstanceTopologyRef = useRef(applyInstanceTopology)
   applyInstanceTopologyRef.current = applyInstanceTopology
```
misi-bp: I'm not sure about what the proper solution should be here, but my brain has issues differentiating 

refetchAndApplyTopology
refetchAndApplyAgenticTopology
applyInstanceTopology
applyInstanceTopologyRef

Can you please rename them to have more descriptive and differentiating var names?

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
```diff
@@ -286,114 +278,36 @@ export function useWorkflowGraphFromAgenticApi({
...
 
-  const scheduleTopologyRefetchRef = useRef(scheduleTopologyRefetch)
   scheduleTopologyRefetchRef.current = scheduleTopologyRefetch
 
   const handleWorkflowInstanceSseEvent = useCallback(
     (ev: EventV1Wire, catalogWorkflowName: string, instanceId: string) => {
-      if (!eventTouchesInstance(ev, catalogWorkflowName, instanceId)) return
-
-      const partial = extractInstanceTopologyFromEvent(
+      const refs = agenticSessionRefs.current
```
misi-bp: Also, this ref object does not contain a list, therefore plural is not a proper name for it

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
```diff
@@ -413,91 +327,24 @@ export function useWorkflowGraphFromAgenticApi({
...
-        if (cancelled) return
-        attachSse()
-      } catch (e) {
-        if (!cancelled) {
-          const msg = e instanceof Error ? e.message : String(e)
-          setAgenticError(msg)
-          logger.apiError("agentic-workflows/bootstrap", e)
-        }
-      }
+      await bootstrapAgenticWorkflowGraph({
```
misi-bp: AFAIK If there's no return value of an async function's call, await is obsolete. If there's no await, async is unnecessary. Calling an async function in a sync function is totally fine, but if it doesn't return anything, it doesn't need 'await' before it and async when defining the sync function itself.
```
const anyFunction = async (prop) => { // <-- async is obsolete, because none of the jobs below is ASYNC AND RETURNS a value
  // do some jobs, like 
  // const randomVar = doSomething(prop) // IF this function is sync
  await fetchSomeData // <-- await is obsolete, since it doesn't return anything
}  
```

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/workflowGraphAgenticInstanceSession.ts
```diff
@@ -0,0 +1,285 @@
...
+    )
+    if (isCancelled()) return
+    const pathUuid = instanceIdToPathUuid(instanceId)
+    const inst = await getWorkflowInstanceState(
+      client,
+      catalogWorkflowName,
+      pathUuid,
+      true,
+    )
+    if (isCancelled()) return
```
misi-bp: I haven't seen anything like this in JS why isCancelled should be checked this many times. If a param (isCancelled) changes, it means the whole function will run again, and everything that is not stored in memory will be calculated again. So if isCancelled is checked once, it should be enough, and when isCancelled changes, it means the function should be rerun again. I can imagine edge-cases though...

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
delthazor: I may be wrong here, but as I understand, the umbrella does not duplicate state. It holds stable RefObject identities so pure helpers in another file can do calls like `refs.applyInstanceTopologyRef.current(...)` without importing the hook. It could be refactored to use a single object, like `useRef<AgenticSessionContext>` but it would be a much larger change, because callbacks would need a fresh object on each render anyways. This change is getting quite far from the original topic anyways, so I would skip that for now.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
delthazor: You are right, I will look into renaming those.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
delthazor: await in this case is used for error propagation (and completion ordering). You are right that it could be more significantly signaling intent with a try-catch for that matter, but I found it less verbose to use it this way.

Do you insist on changing this? I could go with either style, but would rather not touch if not necessary.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/workflowGraphAgenticInstanceSession.ts
delthazor: When the user changes workflow or unmounts, cleanup sets `cancelled = true` but does not restart bootstrap. The in-flight async function from the previous effect keeps running until the next await boundary. To my knowledge, that is standard React effect cancellation.
So even if the function is later re-runs because of the changed value, the previous async call might access the new value in the middle of the execution. It might be an edge case, as you mentioned, but unless I'm wrong about how async effect cancellation is handled in React and/or JS (which can be the case, because JS is black magic to me, and React is like a master black magician building a secret magic tower), I would keep the checks in place.
However I could add a wrapper that helps readability because I agree that it looks awful.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
misi-bp: Using try-catch _inside_ the 'bootstrapAgenticWorkflowGraph' function is enough, therefore no need to use async and await in this function. Please try it, my assumption is no call order will change. If any other issue emerges, please let me know.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
delthazor: Renamed

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
delthazor: Renamed

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useWorkflowGraphFromAgenticApi.ts
delthazor: Changed to try-catch

misi-bp: That was not what I meant, but I see this code snippet is outdated, let's close this.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/workflowGraphAgenticInstanceSession.ts
misi-bp: I investigated this solution, and despite its awful look, for special cases like the one above, this solution is totally fine. But this one is also outdated, let's close it.

misi-bp:
I'm not sure if it's intentional, but with my updates on the main branch, directoryAgentSlugFromAgentRecordUri is not on the live code path today.

The function is only called from topologyWireToReactFlow:

topologyToReactFlow.tsx
Lines 223-225

```
    const directoryAgentSlug = directoryAgentSlugFromAgentRecordUri(
      n.agent_record_uri as string | undefined,
    )
```

But when agentic mode applies topology, it returns early if a static ID map exists:


useWorkflowGraphTopologySync.ts
Lines 48-56

```
  const applyInstanceTopology = useCallback(
    (topology: TopologyWire | undefined) => {
      if (staticIdMapRef.current) {
        onAppliedRef.current?.()
        queueMicrotask(() => {
          restoreEdgeAnimation()
        })
        return
      }
      
```

All four current patterns have a static map:


topologyStaticIdMap.ts
Lines 115-128

```
export function staticIdMapForPattern(
  pattern: PatternType,
): StaticIdMap | null {
  switch (pattern) {
    case PATTERNS.PUBLISH_SUBSCRIBE:
    case PATTERNS.PUBLISH_SUBSCRIBE_STREAMING:
      return buildPublishSubscribeIdMap()
    case PATTERNS.GROUP_MESSAGING:
      return buildGroupCommunicationIdMap()
    case PATTERNS.A2A_HTTP:
      return buildDiscoveryIdMap()
    default:
      return null
  }
}
```

So if we want to introduce this now, the static mapping has to be removed together with the early return.

misi-bp (changes requested):

delthazor:
> So if we want to introduce this now, the static mapping has to be removed together with the early return.

Perfectly right.
I originally planned to do this in the follow-up PR, but that had so much in scope that this was forgotten and left-out. I will resolve this here then.

delthazor:
Sorry, I thought I can managed this without full on attacking the dynamic graph drawing issue, as I wanted that in a separate PR, but I cannot do that without causing a mess.

I know it will greatly expand beyond the scope of this PR, but I have to re-do the backend-driven dynamic graph drawing to be able to remove those static maps that short circuit that call now.

misi-bp:
I am sorry, I was not aware of a following PR - If we have a ticket describing that as well, and we can refer to it from here, I'm more than happy to approve this way.

delthazor:
I change this into a draft until rework is done on other PRs.
The requested change will probably get a separate PR also.

delthazor:
Rescoped PR after graph render/drawing changes were merged. Now its scope is once again just the dynamic URL derivation.
Other requested static mappings and early-return paths were taken care of in previous PRs regarding the backend driven graph-rendering solution.
Some leftover static mappings are still pending, they will be handled in this PR:
https://github.com/agntcy/coffeeAgntcy/pull/650
(although, probably it will be split into 2-3 smaller PRs for better review-ability)

pregnor (approved):
LGTM, get @misi-bp 's approval

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/topologyToReactFlow.test.tsx
misi-bp, line 403: In this case, I suggest using an eslint-disable-<too-many-lines-rule> in this file because this is not something it makes sense to be split, but also producing warnings is not OK.
Thank you

misi-bp (dismissed):

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/topologyToReactFlow.test.tsx
delthazor: I marked the file as an exception. Thanks for the input.

pregnor (dismissed):

misi-bp (approved):

pregnor (approved):

PR #596, opened 2026-05-27
feat: Dockerfiles - pin uv and npm cache location and use per-project id field in Buildkit cache

### LINKED ISSUE #593: [Feature]: Tweak Docker builds after starting from uv base image
### Problem Statement

2 follow up ideas for https://github.com/agntcy/coffeeAgntcy/pull/581 :
1. be explicit about where the `uv` cache is expected to be found
    * thread: https://github.com/agntcy/coffeeAgntcy/pull/581#discussion_r3303768175
2. Seeing as the dependencies may not always be identical between our subprojects (corto, lungo, recruiter), it would be better to `id` the `uv` cache backing folder per project
    * thread: https://github.com/agntcy/coffeeAgntcy/pull/581#discussion_r3303806753
3. similar ideas but for `npm` for the frontend Dockerfiles

### Proposed Solution

1. agreed upon approach is with `ENV UV_CACHE_DIR` statements
2. agreed to do an id that is analogous to `--mount=type=cache,id=recruiter-uv,target=/root/.cache/uv`
3. analogous solutions for `npm` using `npm_config_cache` and `id=<prj>-npm`

### Alternatives Considered

_No response_

### How uv decides where its cache lives

`uv` resolves its cache directory in this order of precedence (highest first):
  * `--cache-dir <path>` CLI flag — if passed, this wins.
  * `UV_CACHE_DIR` environment variable — if set, this is used.
  * Platform default, derived from XDG / OS conventions:
    * Linux: `$XDG_CACHE_HOME/uv`, falling back to $HOME/.cache/uv when XDG_CACHE_HOME is unset.
    * macOS: `~/Library/Caches/uv`.
    * Windows: `%LOCALAPPDATA%\uv\cache`.

For example, In the Recruiter Dockerfile, none of `--cache-dir`, `UV_CACHE_DIR`, or `XDG_CACHE_HOME` is set. The image runs as `root`, so `$HOME = /root`, and the Linux default becomes: `$HOME/.cache/uv  →  /root/.cache/uv`

### How BuildKit picks which host-side cache to attach

The mapping is controlled by the mount's `id`:
  * If you write `--mount=type=cache,target=/root/.cache/uv`, BuildKit uses the target value as the implicit `id`. So every `RUN` in every Dockerfile on that builder that says `target=/root/.cache/uv` ends up sharing the same host-side cache directory.
  * If you want isolation (say, a separate `uv` cache per project), use an explicit `id`: 
    * `RUN --mount=type=cache,id=recruiter-uv,target=/root/.cache/uv ...`
    * Now the host-side store keyed by `recruiter-uv` is what gets mounted at `/root/.cache/uv` inside the build step. 

Two different Dockerfiles using the same `id` will share; using different ids will not.

### Checklist

mihaialexandrescu:
# Description

This PR aims to address 2 follow up ideas for https://github.com/agntcy/coffeeAgntcy/pull/581 :
  1. be explicit about where the uv cache is expected to be found
    * thread: https://github.com/agntcy/coffeeAgntcy/pull/581#discussion_r3303768175
  2. Seeing as the dependencies may not always be identical between our subprojects (corto, lungo, recruiter), it would be better to id the `uv` cache backing folder per project
    * thread: https://github.com/agntcy/coffeeAgntcy/pull/581#discussion_r3303806753
  3. analogous solutions but for npm (using `npm_config_cache` and `id=<prj>-npm`)

More context and details can be found in Issue 593 (also linked below).

## Issue Link

Fixes #593 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

---

Why `npm_config_cache` and not `NPM_CONFIG_CACHE`. `npm` reads config from env vars by lowercasing the variable name and stripping the `npm_config_` prefix (so `npm_config_cache` → the `cache` config key). The uppercase form also works, but lowercase is the documented npm-native convention and matches what the npm docs show. Either would have been fine; I went with the canonical npm form.

### DIFF
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.exchange
@@ -21,8 +21,11 @@ WORKDIR /app
 # Silences warnings about not being able to link files since the cache and sync target are on separate file systems
 ENV UV_LINK_MODE=copy
 
+# Pin the uv cache location so BuildKit cache mounts and uv agree on the path
+ENV UV_CACHE_DIR=/root/.cache/uv
+
 # Install dependencies
-RUN --mount=type=cache,target=/root/.cache/uv \
+RUN --mount=type=cache,id=corto-uv,target=/root/.cache/uv \
     --mount=type=bind,source=coffeeAGNTCY/coffee_agents/corto/uv.lock,target=uv.lock \
     --mount=type=bind,source=coffeeAGNTCY/coffee_agents/corto/pyproject.toml,target=pyproject.toml \
     uv sync --locked --no-install-project
@@ -30,7 +33,7 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 COPY coffeeAGNTCY/coffee_agents/corto/ .
 
 # Sync the project
-RUN --mount=type=cache,target=/root/.cache/uv \
+RUN --mount=type=cache,id=corto-uv,target=/root/.cache/uv \
     uv sync --locked
 
 ## Generate about.properties file with build metadata
--- a/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
+++ b/coffeeAGNTCY/coffee_agents/corto/docker/Dockerfile.farm
@@ -6,8 +6,11 @@ WORKDIR /app
 # Silences warnings about not being able to link files since the cache and sync target are on separate file systems
 ENV UV_LINK_MODE=copy
 
+# Pin the uv cache location so BuildKit cache mounts and uv agree on the path
+ENV UV_CACHE_DIR=/root/.cache/uv
+
 # Install dependencies
-RUN --mount=type=cache,target=/root/.cache/uv \
+RUN --mount=type=cache,id=corto-uv,target=/root/.cache/uv \
     --mount=type=bind,source=coffeeAGNTCY/coffee_agents/corto/uv.lock,target=uv.lock \
     --mount=type=bind,source=coffeeAGNTCY/coffee_agents/corto/pyproject.toml,target=pyproject.toml \
     uv sync --locked --no-install-project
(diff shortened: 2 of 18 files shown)


mihaialexandrescu:
One caveat flagged by Cursor for the UI Dockerfiles:

**Cache scope vs. effectiveness**
Each UI Dockerfile only has a single `npm ci` step with a cache mount. The follow-up `npm install --verbose` and `npm run build --verbose` on lines 15–16 run after `COPY … .` without any cache mount, so they don't benefit from `/root/.npm` at all.   
That's pre-existing behavior and could be considered out of scope for issue #593, but if we wanted, those two `RUN`s could also get `--mount=type=cache,id=<project>-npm,target=/root/.npm` for full reuse across both steps in a single project. I didn't change those because they weren't in the cache-id-isolation discussion thread.

@misi-bp, what do you think about it ?

Later edit:
I created #598 to continue the discussion on it and find it out if the idea holds water.

pregnor:
> One caveat flagged by Cursor for the UI Dockerfiles:
> 
> **Cache scope vs. effectiveness** Each UI Dockerfile only has a single `npm ci` step with a cache mount. The follow-up `npm install --verbose` and `npm run build --verbose` on lines 15–16 run after `COPY … .` without any cache mount, so they don't benefit from `/root/.npm` at all. That's pre-existing behavior and could be considered out of scope for issue #593, but if we wanted, those two `RUN`s could also get `--mount=type=cache,id=<project>-npm,target=/root/.npm` for full reuse across both steps in a single project. I didn't change those because they weren't in the cache-id-isolation discussion thread.
> 
> @misi-bp, what do you think about it ?

I'm all for it if you agree, fine to have it separately.

pregnor (approved):
LGTM

delthazor (approved):
lgtm

PR #616, opened 2026-06-04
feat(lungo): move otel-collector to default docker compose profile

### LINKED ISSUE #610: Make otel-collector a default Docker service
`otel-collector` as a default docker service — promote the collector out of the `observability` profile so basic tracing/event ingestion works without enabling the full stack.

### otel-collector default service

Move `otel-collector` from the `observability` profile to a default service in the main docker compose.
The full observability profile (grafana / tempo / etc.) remains gated by the profile; this change **only promotes the collector itself**.
Default config for the collector must keep working when the rest of the observability stack is not running (e.g. no required exporters that point at services brought up only by the profile, or those exporters must tolerate the targets being down).
The Lungo BE event/tracing path that points at the collector must function in the default compose.

mihaialexandrescu:
# Description

The main goal of the task is to move the `otel-collector` to the default profile and remove its dependency on Clickhouse.

The outline of the proposed solution is to:
  * mount a whole folder with various OTEL config files versions and dependencies, instead of a single file
  * use env vars (specified and explained in `.env`) to determine which config files get used and composed
    * these env vars are referenced directly inside the `otelcol-contrib`/`otelcol` application config config.
  * the glue comes from the combination of these `otelcol-contrib`/`otelcol` application config options: `${file:...}` and `${env:...:-default}`  resulting in  `${file:${env:VAR:-default}}`
  * By default we point to the `nop` configs and if we have Clickhouse then we should uncomment the lines about `exporters` and `pipelines` mentioned in the `.env.example` file.

One upside of this solution is that it makes configuration a good deal more modular, more flexible and ready to mimic the notion of profiles for various environments.

The way this is going to work when Clickhouse is deployed too is that `otel-collector` will deploy and restart a few times (2 in my testing) until the Clickhouse instance is up and the connection can be established.

The additions of the `health_check` extension and the `retry_on_failure` exporter config are just nice-to-haves in my opinion.  
We can't actually use the `health_check` for much because the container is distroless so it doesn't have `curl`/`wget`/etc in it but if we were to use it from the host we'd see a result like this:
```
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 04 Jun 2026 12:34:26 GMT
< Content-Length: 99
< Connection: close
<
* Closing connection
{"status":"Server available","upSince":"2026-06-04T12:30:40.821169964Z","uptime":"3m45.830804562s"}
```

### Problems to circumvent or that excluded other approaches

* If we keep the `clickhouse` `exporters` and `service.pipelines` configurations without having Clickhouse deployed, then the `lungo-otel-collector` keeps restarting and the same logs sequence fills the screen.
  ```
  lungo-otel-collector   | 2026-06-03T08:50:14.774Z	info	otelconftelemetry/tracer.go:47	Internal trace telemetry disabled	{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}}  l View Logs   w Enable Watch   d Detach
  lungo-otel-collector   | 2026-06-03T08:50:14.776Z	info	service@v0.153.0/service.go:241	Starting otelcol-contrib...	{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}, "Version": "0.153.0", "NumCPU": 8}
  lungo-otel-collector   | 2026-06-03T08:50:14.776Z	info	extensions/extensions.go:41	Starting extensions...	{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}}
  lungo-otel-collector   | 2026-06-03T08:50:14.786Z	error	graph/graph.go:437	Failed to start component	{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}, "error": "exec create traces table sql: dial tcp: lookup clickhouse-server on 127.0.0.11:53: no such host", "type": "Exporter", "id": "clickhouse"}
  lungo-otel-collector   | 2026-06-03T08:50:14.786Z	info	service@v0.153.0/service.go:278	Starting shutdown...{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}}
  lungo-otel-collector   | 2026-06-03T08:50:14.786Z	info	extensions/extensions.go:69	Stopping extensions...	{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}}
  lungo-otel-collector   | 2026-06-03T08:50:14.786Z	info	service@v0.153.0/service.go:292	Shutdown complete.	{"resource": {"service.instance.id": "eb52c3ab-10c9-412b-ac38-69cfeec2da60", "service.name": "otelcol-contrib", "service.version": "0.153.0"}}
  lungo-otel-collector   | Error: cannot start pipelines: failed to start "clickhouse" exporter: exec create traces table sql: dial tcp: lookup clickhouse-server on 127.0.0.11:53: no such host
  lungo-otel-collector exited with code 1 (restarting)
  ```
  * This stays true even if we configure the `retry_on_failure` set of options(https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter and in general https://github.com/open-telemetry/opentelemetry-collector/tree/v0.153.0/exporter/exporterhelper). 
  * when we don't have Clickhouse deployed, I think we should use, for now, the `nop` exporter (https://github.com/open-telemetry/opentelemetry-collector/tree/v0.153.0/exporter/nopexporter) to avoid these constant crashes/restarts.

* We must have at least one exporter configured (we can't have an empty `exporters:`), otherwise we also get crashes/restarts:
  ```
  lungo-otel-collector  | Error: invalid configuration: no exporter configuration specified in config
  lungo-otel-collector exited with code 1 (restarting)
  ```

* The `otel/opentelemetry-collector-contrib` image is distroless.
  * That means it doesn't have any shell bundled inside of it. That is a problem if we'd intend to do use environment variables on the command line inside the container because variable expansion is a function of the shell (which we don't have).
  * The only way to use env vars inside the container is to have the application itself use them with its own mechanisms. Nothing the shells do can be used.

* Keep in mind that:
  > The `service` section is used to configure what components are enabled in the Collector based on the configuration found in the `receivers`, `processors`, `exporters`, and `extensions` sections. If a component is configured, but not defined within the `service` section, **then it’s not enabled**. 

* The "Override Settings" mechanism (https://opentelemetry.io/docs/collector/configuration/#override-settings) unfortunately only merges the settings overriden from the CLI.
  * I tested with syntax like `--set exporters::nop={} ` and `--set exporters={} --set exporters::nop={} ` but they didn't remove the `clickhouse` exporter. This limits heavily the usefulness of this feature for the purposes of this task.

* OTel's `${env:...}` **only does variable substitution and default-fallback substitution**; it **can't test** whether that string contains `observability` (if we were to try to read and test the contents of the `COMPOSE_PROFILES` env var).

* Something like this is `docker-compose.yaml` would rely on the same env vars existing on the host (not in the container) which is not useful when just calling `docker compose`.
  ```yaml
      env_file:
        - .env
      environment:
        - OTELCOL_EXPORTERS_CFG_FILE=${OTELCOL_EXPORTERS_CFG_FILE:-/etc/otel/cfg_exporters-1-nop.yaml}
        - OTELCOL_PIPELINES_CFG_FILE=${OTELCOL_PIPELINES_CFG_FILE:-/etc/otel/cfg_pipelines-1-nop.yaml}
  ```
  * If we were to generally use Makefile targets or Taskfile tasks, then we could consider such an option but with our current workflow we can't.

## Issue Link

Fixes #610 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


delthazor (dismissed):
lgtm

coffeeAGNTCY/coffee_agents/lungo/docker-compose.yaml
```diff
@@ -265,14 +265,12 @@ services:
     ports:
       - "4317:4317"
       - "4318:4318"
+      - "13133:13133"
```
pregnor, line 268: **Question**: What do we need the new port for?

pregnor (dismissed):

pregnor (commented):
One last thing that came to me: could you do a run through the docs to update the observability compose profile and mark it optional if we ever called it required?

coffeeAGNTCY/coffee_agents/lungo/docker-compose.yaml
mihaialexandrescu: From the `health_check` extension (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/healthcheckextension/README.md) I enabled in `coffeeAGNTCY/coffee_agents/lungo/config/docker/otel/otel-collector-config.yaml` (just thought it'd be a good idea):
```
extensions:
  health_check:
    endpoint: 0.0.0.0:13133
    path: /health/status
```

We can't actually use the `health_check` for much because the container is distroless so it doesn't have `curl`/`wget`/etc in it but if we were to use it from the host (with a `curl`) we'd see a result like this:
```
< HTTP/1.1 200 OK
< Content-Type: application/json
...
<
* Closing connection
{"status":"Server available","upSince":"2026-06-04T12:30:40.821169964Z","uptime":"3m45.830804562s"}
```

pregnor (approved):

mihaialexandrescu:
> One last thing that came to me: could you do a run through the docs to update the observability compose profile and mark it optional if we ever called it required?

Done in https://github.com/agntcy/coffeeAgntcy/commit/1deb97f9f71595a3cf44c48a56032b168bc22677.

delthazor (approved):
lgmt

Issue #587, opened 2026-05-27
[Feat] Fiber: MCP tool-call event emission


codyhartsook:
Parent: #592

## Summary

Add a **client-side MCP interceptor** that emits workflow events for every MCP tool call. The interceptor follows the same shape as the existing A2A client middleware (`common/a2a_event_middleware/middleware.py`): it intercepts outbound calls, builds an `Event` conforming to the shared workflow event schema (`schema/jsonschemas/event_v1.json`), and publishes it via `WorkflowAPIEventSink` directly to the workflow API. **Events do not travel over A2A transport.**

To support this without duplication, the reusable parts of `common/a2a_event_middleware/` (sink, in-flight/trace helpers, workflow catalog, event/topology builders) are extracted into a transport-agnostic location. `common/a2a_event_middleware/middleware.py` remains the only A2A-specific file.

Each MCP tool call is modeled as a **transient topology node** in the workflow instance: a node (and edge from the invoking agent) is created (`operation: create`) on tool start and removed (`operation: delete`) on tool end. This gives the frontend a live picture of in-flight tool calls without inventing a new event type or extending `event_v1`.

**Scope:** Backend only. UI consumes whatever lands on the existing SSE stream.

## Goals

- Every MCP tool call originating from a coffeeAgntcy agent produces a start and end `Event` posted to the workflow API.
- Events conform to `event_v1` and reuse the existing in-memory store and SSE path established by #430.
- No code duplicated between A2A and MCP event emission: the sink, trace/correlation helpers, workflow catalog, and event builders live in a transport-agnostic module imported by both.
- Trace correlation preserved via the standard helpers in `inflight.py`.
- No tool arguments or results in event payloads.

### Interceptor location

- **Client-side only.** Wrap the MCP client used by Lungo agents (reference: `coffeeAGNTCY/coffee_agents/lungo/agents/farms/colombia/agent.py`).
- Server-side wrapping (`coffee_agents/lungo/agents/mcp_servers/`) is **out of scope** for fiber.

### Schema

- Reuse `schema/jsonschemas/event_v1.json` unchanged. An MCP tool call is expressed as a transient topology change on the workflow instance:
  - **Tool start:** one node with `operation: create` (type `mcp_tool_call`, stable per-call `node_id`), plus one edge with `operation: create` from the invoking agent's node to the tool-call node.
  - **Tool end:** the same node and edge with `operation: delete`.
[...]

PR #630, opened 2026-06-09
(feat) Implement MCP live event processing

delthazor:
# Description

MCP call_tool invocations now emit event_v1 topology updates (CREATE on start, DELETE on end) to the workflow API, reusing the shared workflow_utils builders/sink and the existing SSE ingestion path.

Each tool call is modeled as a transient mcp_tool_call node linked from the invoking agent. The wrapper handles errors and streamed results, refcounts overlapping calls, and removes the MCP-emitted agent node when the last call finishes. Workflow identity is propagated from the auction supervisor to the Colombia farm via A2A message metadata and OTel baggage. Weather and payment MCP call sites are wired, unit tests and middleware docs are included.

### Improtant:
The change scope is backend only! 
No new frontend node type or other change is introduced to make special rendering for MCP. (This is how the issue stated its scope)

## Issue Link

Implements #587 

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/common/workflow_utils/mcp.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/common/workflow_utils/mcp.py
@@ -1,0 +1,176 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+"""Transport-agnostic builders for MCP tool-call topology events.
+
+An MCP tool call is modeled as a transient topology node: a CREATE event when
+the tool call starts and a DELETE event when it ends (success or error). The
+invoking agent node carries a stable_agent_id so it merges with the same agent
+node the A2A middleware emits.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from schema.types import (
+	Event,
+	EventType,
+	Operation,
+	PartialTopology,
+	TopologyEdgeItem,
+	TopologyNodeItem,
+)
+
+from common.stable_agent_id import stable_agent_id_for_name
+from common.workflow_utils.builders import build_event, make_edge, make_node
+from common.workflow_utils.event_sink import EventSink
+from common.workflow_utils.inflight import RuntimeIdAllocator
+
+MCP_TOOL_NODE_TYPE = "mcp_tool_call"
+_AGENT_NODE_TYPE = "customNode"
+
+
+async def build_mcp_tool_topology(
+	agent_id: str,
+	tool_name: str,
+	mcp_server: str,
+	*,
+	operation: Operation,
+	allocator: RuntimeIdAllocator,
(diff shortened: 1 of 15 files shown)


coffeeAGNTCY/coffee_agents/lungo/common/workflow_utils/mcp.py
```diff
@@ -0,0 +1,176 @@
...
+	if operation == Operation.CREATE:
+		nodes: list[TopologyNodeItem] = [
+			make_node(
+				agent_node,
+				operation=Operation.CREATE,
+				node_type=_AGENT_NODE_TYPE,
+				label=agent_id,
+				layer_index=layer_index,
+				stable_agent_id=stable_agent_id_for_name(agent_id),
+			),
```
pregnor, line 67: **Question**: Why are we creating the agent node here, not updating the existing one?
Are we handling the agent-mcp tool call as a separate graph here? If so why?

pregnor (approved):
LGTM in general.

1 Q regarding design choices.

Get @codyhartsook 's approval as well before merging.

coffeeAGNTCY/coffee_agents/lungo/common/workflow_utils/mcp.py
delthazor: This lands in the same graph due to stable agent id, but it is on the UI to collapse it instead of duplicating.

We CREATE the agent node (with stable_agent_id) instead of UPDATE because the farm process doesn’t know the A2A middleware’s node:// id for that agent. The UPDATE would be ignored by the merge layer. This matches outbound A2A, which also CREATEs agent nodes with stable_agent_id.

The agent node representing the MCP is a transient anchor for the 'agent → mcp_tool_call' edge and is deleted when the last MCP call finishes. The UI collapses it with the existing A2A farm node via stable_agent_id, and the durable node is the one from the A2A path.

Of course this may need an update when UI refactor is done, despite that the PR does not contain direct UI changes.

pregnor: Eh, seems a bit hacky to me, would be nicer if we propagated the missing info here.

But it can fly for today until I come up with an exact solution to this.

delthazor: In that case feel free to open an issue for the correction, and assign it to me.

pregnor: https://github.com/agntcy/coffeeAgntcy/issues/649

codyhartsook (approved):
lgtm

PR #641, opened 2026-06-12
ci(gha,docker,helm): add version guard.

### LINKED ISSUE #637: Fix CI workflows not to overwrite/overpush existing tags (helm, docker)
Currently we are overpushing without checking.

pregnor:
# Description

Before: 
- We are not tagging helm charts in Git, we are using the Chart.yaml version for tagging, this is preexisting to my arrival and is a suggested practice to avoid unnecessary work, but with the current tag handling it needed a bit of safeguarding on other events to avoid overpushing preexisting chart versions.
- We allowed overpushing charts (and maybe images as well, but I'm not 100% sure if that would require an extra flag).

After:
- Implemented a workflow_dispatch manual workflow to be able to delete ghcr.io package versions.
- Added a guard to the docker build and push workflow that checks whether any image we would push in a push scenario exists already, collects all those and errors the workflow without proceeding to push.
- Added a guard to the helm package and push workflow that checks whether any chart we would push in a push scenario exists already, collects all those and errors the workflow without proceeding to push.

Consequences:
- Chart release afterwards happens whenever we merge chart.yaml version to main. In the past we did this + on each tag we overpushed the existing chart from the tag ref that broke immutability - and caused ArgoCD issues.
- This also enforces the current unwritten practice of bumping chart versions in each PR we modify charts in.
- There is another guard added to not care about unmodified charts when checking the preexisting versions.

## Issue Link

fixes #637 

## Type of Change

- [X] Other (please describe): CI

## Checklist

### DIFF
--- a/.github/workflows/delete-ghcr-package-version.yaml
+++ b/.github/workflows/delete-ghcr-package-version.yaml
@@ -1,0 +1,75 @@
+name: Delete GHCR Package Version
+
+on:
+  workflow_dispatch:
+    inputs:
+      package_type:
+        description: "GitHub Packages type (ghcr.io uses container)"
+        type: choice
+        required: true
+        default: container
+        options: [container]
+      owner:
+        description: "Org or user that owns the package"
+        type: choice
+        required: true
+        default: agntcy
+        options: [agntcy]
+      repository:
+        description: "Repository that contains the package"
+        type: choice
+        required: true
+        default: coffee-agntcy
+        options: [coffee-agntcy]
+      package_name:
+        description: "Package path without ghcr.io/owner/repository (e.g. lungo-ui or helm/lungo-ui)"
+        type: string
+        required: true
+      version:
+        description: "Tag or version to delete (e.g. latest, 1.2.3, pr-42)"
+        type: string
+        required: true
+
+jobs:
+  delete:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    steps:
+      - name: Delete package version if it exists
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+
+          package_type=${{ inputs.package_type }}
+          owner=${{ inputs.owner }}
+          repository=${{ inputs.repository }}
+          package_name=${{ inputs.package_name }}
+          version=${{ inputs.version }}
+
+          if echo "${package_name}" | grep -q '^helm/'; then
+            repository="coffee_agntcy" # Note: Legacy naming uses underscore instead of hyphen.
+          fi
+
+          echo "Deleting package ${package_type} version: ghcr.io/${owner}/${repository}/${package_name}:${version}"
+
+          package_encoded=$(jq -nr --arg p "${repository}/${package_name}" '$p|@uri')
+          api_base="/orgs/${owner}/packages/${package_type}/${package_encoded}"
+
+          # Note: name below matches the digest for images/charts.
+          version_id=$(gh api "${api_base}/versions" --paginate \
+            --jq ".[] | select(
+              .name == \"${version}\" or
+              (.metadata.container.tags[]? == \"${version}\")
+            ) | .id" \
+            | head -n1)
+
+          if [[ -z "${version_id}" ]]; then
+            echo "Not found: ghcr.io/${owner}/${repository}/${package_name}:${version}"
+            exit 0
+          fi
+
+          gh api -X DELETE "${api_base}/versions/${version_id}"
+          echo "Deleted: ghcr.io/${owner}/${repository}/${package_name}:${version} (id ${version_id})"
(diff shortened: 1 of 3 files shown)


codyhartsook (dismissed):
lgtm

pregnor:
I realized we would still be pushing charts that have no changes and no version change either, while erroring on pushing charts without versions changes.

I want to fix the first because I know the Helm package on the same content in 2 different time moments are not going to be equivalent, so this is still not good enough.

I might refactor how we fire off our matrices in order to prevent these dummy pushes altogether, but that requires a bit more time for me.

delthazor:
I'm not really knowledgeable in this topic, so here are some dummy questions (feel free to ignore them, if they do not apply):
-Are these properly wired in? Should they run in-parallel to the build jobs or block them? Is the job dependency correct?
-Does Helm integrity check needs a checkout? It looks like using a git diff, but I'm not sure if at that point the file is reachable or not. Is that git diff safe to use?
-How package jobs handle the push flag compared to the subproject-to-build? For example a tag push would be handled as a GitHub event "push" or a "workflow dispatch"? Is the push flag ignored in that case? Does that cause an overpush loophole around your fix?

pregnor:
> I'm not really knowledgeable in this topic, so here are some dummy questions (feel free to ignore them, if they do not apply): -Are these properly wired in? Should they run in-parallel to the build jobs or block them? Is the job dependency correct? -Does Helm integrity check needs a checkout? It looks like using a git diff, but I'm not sure if at that point the file is reachable or not. Is that git diff safe to use? -How package jobs handle the push flag compared to the subproject-to-build? For example a tag push would be handled as a GitHub event "push" or a "workflow dispatch"? Is the push flag ignored in that case? Does that cause an overpush loophole around your fix?

1. Helm push workflow is run parallel to Docker build and Integration test workflows.
    Helm package workflow - the spawns of helm push - runs parallel to its own siblings.
    They are wired in this way.
    My new job is properly depends on the inputs it needs, but there was an issue here with it being a dependency of the followings, thanks for noticing, fixed in [151bc62](https://github.com/agntcy/coffeeAgntcy/pull/641/commits/151bc62ca31ee46745225e77625b38c779efd196).

2. Absolutely, there was a missing checkout there, thanks for catching.
    (I had to double check myself.)
    Fixed in [8d0e695](https://github.com/agntcy/coffeeAgntcy/pull/641/commits/8d0e69590cd8d54944ba0266c235bfdc9cbfcd20).

3. Tag push is a push event with the tag as the ref name. Only manual runs are workflow dispatches.
    Let me break it down how it works in each cases:

Docker:

- `docker-build-push.yaml`:23 `push` default is false for `workflow_dispatch`, but it will be set to the manual input the initiator provides there.
- in -||-:169, `compute-build-parameters`/`build_params` we translate this into 

    - explicit local `push=false` for `pull_request` at 177 (made it explicit just now in [0a1c8bf](https://github.com/agntcy/coffeeAgntcy/pull/641/commits/0a1c8bf2539b70370c5ce4b3697a2377453a0cdf)),
    - explicit local `push=true`for `main` at 181,
    - explicit local `push=true` for `ref/tags` at 184,
    - transitive `push={{ inputs.push }}` value for `workflow_dispatch` at 187

- This is made into an output at -||-:196
- And that output variable is what I reference in `check-push-target-existence` at 213 to set the local push variable to.
- Semantics here is that we push `tag=latest` on main, `tag=git_tag` on tags, nothing on `pull_request` and whatever the user sets in the `image_tag` variable for `workflow_dispatch`.

Helm:

- `helm-push.yaml`:22 `push` default is false for `workflow_dispatch` but it will be set to the manual input the initiator provides there.
- in -||-:99, `subprojects-to-build`/`choose-subprojects` we translate this into

    - explicit local `push=true` for `main` `push` events at 106,
    - explicit local `push=false` for other `push` events at 108,
    - transitive `push={{ inputs.push }}` value for `workflow_dispatch` at 114,
    - explicit `push=false` value for `pull_request` at 118

- This is made into an output at -||-:123,
- And that output variable is what I reference in `check-push-target-existence` at 127 to set the local push variable to.
- Semantics here is that we only push chart versions when `main` is pushed with a chart directory change and we take the version from `Chart.yaml:.version`. Otherwise we let the user set it on `workflow_dispatch` with version values still taken from the `Chart.yaml:.version` and `ref/tags` `push` and `pull_request` does not trigger a chart push.

@mihaialexandrescu please correct me if I misrepresented any of the current logic as you know from before this PR. (the lines changed due to my changes, but the general approach should be the same)

My change is supposed to guard against reusing a chart version.
The missing change that I'm about to do is about making sure we don't unnecessarily trigger chart pushes even when there is no content change (timestamp would change and thus helm tgz would change) we shouldn't even start the push process, that's still to be added.

Edit: fixing github markdown formatting

mihaialexandrescu (dismissed):

mihaialexandrescu (approved):

delthazor (approved):
lgtm

Issue #430, opened 2026-03-11
[Epic] Lungo, BE Live state and A2A events


pregnor:
## Summary

Refactor Lungo backend message flows so the system is **ready to provide live state, topology, and events** to the frontend. 

- The backend will expose 
    - **workflow instance-scoped state**, 
    - **use-case catalog and starting topologies**, and 
    - **SSE for near real-time events**. 
- Agents emit **custom events via A2A**.
- An **A2A middleware** receives these messages and updates the state endpoint (in-memory store) and the state store feeds the SSE. 
- The frontend (in a separate epic) will consume state and SSE. 
- Topology is **workflow instance-scoped** and may evolve during a workflow instance (e.g. recruiter); use-cases define the **starting** graph. 
- Implementation stays close to existing code and uses **open standards** (A2A, HTTP, SSE) and **agentic best practices**.

**Scope:** Backend only. No frontend changes in this epic.

## Goals

- Expose **workflow instance-scoped state** (including current topology) and **SSE** for near real-time events.
- Provide **agentic worfklow API**: use-case list with workflow details, starting topology per use-case, and workflow instance-scoped current state topology (with topology-only filter).
- **A2A middleware for the state endpoint:** state/event updates are received via A2A middleware, which updates the in-memory store that feeds the SSE; agents (and graph runtime) emit custom events via A2A toward this path.
- Keep implementation aligned with current Lungo stack (FastAPI, LangGraph, ioa_observe) and A2A; no new framework lock-in.

## Technical notes (design-agnostic)

- **State store:** In-memory, keyed by workflow instance id. No persistence or replay; current workflow instance only.
- **Event source and state ingestion:** Events (e.g. node enter/exit, handoffs, tool start/end) are emitted via A2A; an **A2A middleware** receives these messages and is responsible for updating the state store which notifies SSE subscribers. This keeps ingestion on a single, standard path and allows multiple agents/services to participate.
- **workflow instance lifecycle:** User selects a **use-case** to start a workflow instance; the workflow instance gets that use-case's **predefined starting topology**. The graph may **evolve** during the workflow instance (e.g. recruiter); all retrieval of "current" state/topology is **workflow instance-scoped**.
- **Real-time:** Frontend will consume via **SSE** (no polling as primary path). Backend must support at least one SSE endpoint per workflow instance. The SSE stream must use **NDJSON** (newline-delimited JSON), with each line based on the same **A2A message schema** defined for the workflow instance state progress payload.
[...]

PR #643, opened 2026-06-13
fix(lungo): workflow event animations for broadcast-type prompts

### LINKED ISSUE #590: [Feat] Fiber: Non-streaming event animation fix
Parent: #592

## Summary

Workflow event animations (driven by the events SSE introduced in #430) don't render in the Lungo UI when the user selects a non-streaming chat endpoint, even though the events SSE is a separate stream and should be independent of chat delivery mode. Root cause is unknown; this epic is **diagnose-first**, then fix on whichever side is broken (FE subscription, BE emission, or both).

**Scope:** Diagnose may touch FE or BE. Fix lives where the bug is.

## Goals

- Identify why workflow event animations stop rendering when the chat endpoint is non-streaming.
- Make animations render identically regardless of chat mode.
- Keep the events SSE as the single source of truth for animation drivers.

## Technical notes

- The events SSE endpoint already exists: `GET /agentic-workflows/{workflow_name}/instances/{workflow_instance_id}/events/stream` (from #430).
- Hypotheses to test during diagnose:
  - **FE:** subscription to the events SSE is gated on the chat-streaming toggle by mistake.
  - **BE:** non-streaming chat path follows a different execution branch that doesn't emit events.
  - **Other:** lifecycle/race (SSE opened too late, closed too early in non-streaming mode), or an instance id mismatch on the non-streaming path.
- Diagnose output is a short root-cause note before any code change.

## Child stories / sub-items

1. **Diagnose** — reproduce the bug, instrument FE subscription + BE event emission, identify root cause. Output: written note pointing to the broken layer.
2. **Fix** — apply the change indicated by (1). Animations driven from the events SSE regardless of chat mode.
3. **Regression coverage** — automated or manual checklist that exercises both streaming and non-streaming chat modes and asserts animations fire in both.

## Acceptance criteria

- Workflow event animations render identically whether the user picks the streaming or the non-streaming chat endpoint.
- Root cause documented (in the issue or a short doc) so the fix is justified, not symptomatic.

mihaialexandrescu:
# Description

Workflow event animations (driven by the events SSE introduced in https://github.com/agntcy/coffeeAgntcy/issues/430) don't render in the Lungo UI when the user selects a non-streaming chat endpoint (e.g. `Supervisor -> Conversation: Purchasing -> Publish Subscribe -> A2A SLIM` menu item).

Note: Animations only work for broadcast/multicast-type communications. That's a limitation and design that sits well outside the scope of the targeted issue. 

The point is that IF we use prompts that go for broadcast/multicast communication (like the first suggested prompt: `What yield do the farms have?`), then this fix works (tested manually) and that's the spirit of the raised GH Issue.

Otherwise (for prompts like: `I want to order 100 lbs of coffee at $5 per pound from the Colombia farm.`) animations don't work anyhow (for neither streaming and non-streaming workflows) which made reproducing the issue in the ticket a lot more time consuming.

### Explanation (AI generated)

The events SSE stream is not gated on the chat-streaming toggle — it opens independently when a catalog workflow is selected. The break is upstream of the SSE, on the chat request path:
  * The frontend instantiates a workflow (`POST /agentic-workflows/{name}/`), gets an `instance://<uuid>`, stores it in `activeWorkflowInstanceStore`, and the events SSE subscribes to that instance id.
  * Streaming chat (`auctionStreamingStore`, `groupStreamingStore`) sends that `workflow_instance_id` in the request body, so backend-emitted workflow events are tagged with the SSE-subscribed instance → animations render.
  * Non-streaming chat (`useAgentAPI.ts` → `POST /agent/prompt`) sent only `{ prompt }`. With no instance id, each supervisor graph mints a throwaway `instance://<uuid4>` that was never registered in the store.
  * The A2A event middleware emits events for that throwaway id. The internal sink POSTs to `.../instances/{minted-uuid}/events/`, which 404s because the instance isn't registered (confirmed by `test_instantiate_delete_then_post_event_returns_404`). The failure is swallowed → no merge → no dispatch → the SSE listener (watching the real instance) receives nothing → `eventTouchesInstance` filters everything → `patchGraphActiveHighlight` never fires → no animation.

The backend handlers already accept workflow_instance_id on both paths.

### Fix (AI generated)

In `coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useAgentAPI.ts`, both non-streaming POSTs (`sendMessage` and `sendMessageWithCallback`) now build their body via a shared `buildPromptRequestBody`, which reads the active instance id from the same store the streaming paths use and includes it when present — mirroring the existing conditional-spread pattern in `auctionStreamingStore`. This makes the backend emit events on the SSE-subscribed instance regardless of chat mode.

## Issue Link

Fixes #590

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


delthazor (dismissed):
seems good

pregnor (changes requested):
LGTM, but I'm blocking in order to avoid having conflicts with the UI refactor, let's merge this in right after, hopefully very soon.

mihaialexandrescu:
In my testing, this problem has not been solved by the UI refactor. This will require a rebase/refactor.

mihaialexandrescu:
I rebased on `main`, retested (ok) and pushed the updates. 
Diff is practically the same.
The solution can be reviewed.

pregnor (approved):
LGTM, but make sure you get an approve from @misi-bp before merging

misi-bp:
@mihaialexandrescu Tested, animation perfectly works at publish subscribe / a2a slim. No errors anywhere. Feel free to merge.

misi-bp (approved):
lgtm

Issue #588, opened 2026-05-27
[Feat] Fiber: Pattern reference library explorer view


codyhartsook:
Parent: #592

## Summary

Make the patterns reference library interactive docs. Patterns are split in the FE sidebar between **implemented patterns** and the **pattern reference library** (docs only).

<img width="1751" height="1294" alt="Image" src="https://github.com/user-attachments/assets/767c1ddd-2c57-4180-977c-a54f6270682a" />

- Clicking an **implemented** pattern continues to work as today: load the real agent services topology in the graph canvas; the existing chat panel talks to the selected agent service's chat endpoint (e.g. `POST /agent/prompt/stream`).
- Clicking a **reference-library** pattern (unimplemented) clears the graph canvas and shows **one large custom "doc node"** containing the pattern's markdown — including a **client-rendered mermaid topology diagram** embedded in that markdown (John provided). The same chat panel is **context-switched** to point at a new `POST /patterns/{name}/chat` endpoint on the workflow api server (plain LLM chat seeded with the pattern markdown).

Backend addition: a new chat endpoint on the workflow api server, server-side LLM call, credentials from the existing Lungo llm config.

**Scope:** Frontend (canvas mode + custom doc node + chat-panel mode switch) and one new backend chat endpoint.

## Goals

- Clicking a pattern in the reference library clears the workflow graph canvas and renders a single large doc node containing the pattern's markdown and its embedded mermaid topology diagram.
- The chat panel, when in pattern-doc mode, sends user turns to the new `POST /patterns/{name}/chat` endpoint and streams the response.
- The chat panel, when in workflow mode (implemented pattern selected), continues to hit the selected agent service's existing chat endpoint with no behaviour change.
- Each entry into pattern-doc mode is a fresh conversation; leaving discards it.

### Canvas modes

- **`workflow` mode** (existing, implemented patterns): real agent services topology in the ReactFlow canvas, driven by #430 state/SSE. No change.
- **`pattern_doc` mode** (new, reference-library patterns): canvas cleared; one custom **`pattern_doc` ReactFlow node** rendered large and fixed, with the markdown content rendered **inside the node body**. No live topology nodes/edges in this mode.

### Custom doc node

- New ReactFlow node type `pattern_doc`. Custom component renders the markdown content inside the node.
- Markdown is rendered FE-side; mermaid fenced blocks (` ```mermaid `) are rendered to SVG client-side as part of normal markdown processing.
[...]

PR #651, opened 2026-06-18
feat(#588): pattern reference library explorer — BE chat + FE canvas

codyhartsook:
## Summary

Implements the pattern reference library explorer (epic #588): clicking a reference-library entry clears the graph, mounts an in-canvas markdown doc surface (with mermaid topology diagrams), and re-points the chat panel at a new docs-grounded LLM endpoint.

Opened as a **draft for review** — combines the backend and frontend slices for the epic.

> **Note:** the bulk mermaid additions to the 25 existing workflow docs were split into **#652** for easier review. This PR keeps only the 3 new pattern-level stub docs (which the chat endpoint depends on).

### Backend
- `POST /patterns/{name}/chat` — NDJSON streaming, docs-grounded ADK advisor with server-side per-session state, `read_pattern_doc` tool, per-pattern isolation, 30-min idle TTL eviction.
- Implementation-assistant **sub-agent** seeded with AGNTCY pillar docs/repos for implementation/repo/demo questions.
- 3 pattern-level doc stubs (Supervisor / Peer Group / Recruiter).

### Frontend
- `pattern_doc` canvas mode: sidebar selection swaps in a `PatternDocCanvas` (markdown + lazy-loaded, theme-aware mermaid) via `CanvasSwitch`.
- `usePatternChatAPI` NDJSON streaming hook + `streamPatternChat` orchestrator; chat panel routes to the pattern endpoint in `pattern_doc` mode with a per-conversation `session_id`.
- Markdown rendering for **all** assistant chat responses (`ChatMarkdown`: react-markdown + remark-gfm + remark-breaks + rehype-sanitize); plain text keeps line breaks, links open in a new tab.
- Composer + message thread widened and centered.

## Testing
- FE: `npm run typecheck && npm run lint:check && npm run test` — green (69 tests).
- BE: pattern-chat unit + live smoke tests (see CLAUDE.md cheat-sheet).

## Related
- #652 — mermaid diagrams for the 25 existing workflow docs (split out of this PR).

Epic: #588


coffeeAGNTCY/coffee_agents/lungo/frontend/eslint.config.mjs
```diff
@@ -92,6 +92,7 @@ export default [
     files: [
```
misi-bp, line 92: @delthazor @codyhartsook 

I am sorry, but I really don't agree with this approach: we have this rule (the max lines) to make the code more readable. This is against it.
I agree with skipping this rule when we have a file with a long mock data, for example.

But adding common components and util stuff to this part is not OK, following this would mean the rule has no point and we can get rid of it.

If you don't agree with my rational, let's discuss it as a team and state our common rules.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/agenticWorkflowsApi.test.ts
```diff
@@ -0,0 +1,90 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import {
+  fetchWorkflowDocumentation,
+  WorkflowDocumentationNotFoundError,
+} from "./agenticWorkflowsApi"
+
+const originalFetch = globalThis.fetch
```
misi-bp, line 12: This is very elegant, reassigning the original fetch function. AI generated or not :)

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/agenticWorkflowsApi.ts
```diff
@@ -142,6 +146,58 @@ export const fetchWorkflowSummariesWithRetry = async (
...
+  title: string
+  full_markdown: string
+}
+
+export const fetchWorkflowDocumentation = async (
+  workflowName: string,
+  signal?: AbortSignal,
+): Promise<WorkflowDocumentation> => {
+  const url = `${getAgenticWorkflowsApiUrl()}/agentic-workflows/${encodeURIComponent(workflowName)}/documentation/`
+  const response = await fetch(url, {
```
misi-bp, line 167: In the near future, I'd like the app to use axios for all the http requests - feel free to leave as-is for now if you don't have the time. Next time please use it to avoid redundant work :)

coffeeAGNTCY/coffee_agents/lungo/frontend/src/useApp.ts
```diff
@@ -48,7 +54,24 @@ export function useApp() {
...
+    string | null
+  >(null)
+  const [patternDocState, setPatternDocState] = useState<PatternDocState>({
+    status: "idle",
+    documentation: null,
+    errorMessage: null,
+  })
+
+  const canvasMode: CanvasMode = useMemo(
+    () => (selectedReferencePattern !== null ? "pattern_doc" : "workflow"),
```
misi-bp: When there's a condition check like this, I suggest using enum instead, so the control values of the checking can't be mistyped --> check the type suggestion. See 'export enum CanvasMode' below.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/types/patternDoc.ts
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Shared types for the reference-library pattern-doc surface.
+ * Placed in a leaf module so MainArea and useApp can both depend on
+ * them without forming a cycle.
+ */
+
+import type { WorkflowDocumentation } from "@/utils/agenticWorkflowsApi"
+
+export type CanvasMode = "workflow" | "pattern_doc"
```
misi-bp: Instead I recommend to use an exported enum for these values --> that also can be used for type checking but it's possible and more safe to reuse any of the exact values of CanvasMode.

```
export enum CanvasMode {
  WORKFLOW = "workflow",
  PATTERN_DOC = "pattern_doc",
}
```

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/agenticWorkflowsApi.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  it("throws WorkflowDocumentationNotFoundError on 404", async () => {
+    mockFetch(404, { detail: "Workflow documentation not found for: MadeUp" })
+
+    await expect(fetchWorkflowDocumentation("MadeUp")).rejects.toBeInstanceOf(
+      WorkflowDocumentationNotFoundError,
+    )
+  })
+
+  it("throws a generic Error on other non-OK responses", async () => {
+    mockFetch(500, { detail: "Server error" })
```
misi-bp, line 85: are 401 and other http response scenarios also handled properly?

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useAppChatState.ts
```diff
@@ -94,15 +99,17 @@ export function useAppChatState({ selectedPattern }: UseAppChatStateParams) {
       setIsAgentLoading(true)
       setButtonClicked(true)
       setApiError(false)
-      if (
+      if (canvasMode === "pattern_doc") {
```
misi-bp: Both conditions let the code run the same command: 
 setShowFinalResponse(true)
Do we need to check the conditions separately? (A simple OR || can get the job done.)

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/PatternDoc/MermaidBlock.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * Copyright AGNTCY Contributors (https://github.com/agntcy)
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useEffect, useId, useRef, useState } from "react"
+import mermaid from "mermaid"
+import { useThemeMode } from "@open-ui-kit/core"
+import { logger } from "@/utils/logger"
+
+const DARK_THEME_VARS = {
```
misi-bp: Please do not use unique colors, especially not this hardcoded way.
Instead, choose the preferred colors and other style settings from here:
https://main--68cc22452afe30d90e4ca977.chromatic.com/?path=/docs/foundations-colors--docs&globals=backgrounds.value:!hex(00142B)

And you can refer to them like this:

````
import { useTheme } from "@open-ui-kit/core"

// ...
  const theme = useTheme()
  const avatarIconColor = theme.palette.grey[50]
// ...

```

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/PatternDoc/PatternDocMarkdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+const isMermaidCodeBlock = (className: string | undefined): boolean =>
+  typeof className === "string" &&
+  className.split(/\s+/).includes("language-mermaid")
+
+const PatternDocMarkdown: React.FC<PatternDocMarkdownProps> = ({
+  markdown,
+  className,
+}) => {
+  return (
+    <div className={className}>
```
misi-bp: This div should be replaced with an OUK Box pls. Also please double-check if each element (outside rendering Mermaid of course) is using OUK components.

misi-bp:
I'm sorry for reviewing this PR in draft mode, I had to do it because of the very close release date. Let me know if any comment of mine is outdated.

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/components/schemas.yaml
```diff
@@ -45,6 +45,22 @@ components:
...
+    PatternChatRequest:
+      type: object
+      required: [session_id, message]
+      additionalProperties: false
+      properties:
+        session_id:
+          type: string
+          minLength: 1
+          maxLength: 128
+          description: Client-minted opaque conversation id (e.g. UUIDv4).
```
pregnor: **Question**: Wouldn't it be more consistent to use the common schema ID format here as well? Is there any downside to it?

```suggestion
      required: [session_id, message]
      additionalProperties: false
      properties:
        session_id:
          type: string
          "pattern": "^session://[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$",
          description: Client-minted opaque conversation id (e.g. UUIDv4).
```

(also in that case I would prop adding it to the schema as a minor change)

If we change this, we also need to change coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/dtos.py line 90 def, the tests, etc.

Because of the release rush again, I can let this fly, but I think we should start thinking on a systematic level about this.

Let me know if I should create a ticket for this.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/pattern_chat.py
```diff
@@ -0,0 +1,246 @@
...
+# (pattern_name::session_id) -> last-used monotonic timestamp.
+_last_used: dict[str, float] = {}
+
+
+_LITELLM_PROXY_BASE_URL = os.getenv("LITELLM_PROXY_BASE_URL")
+_LITELLM_PROXY_API_KEY = os.getenv("LITELLM_PROXY_API_KEY")
+if _LITELLM_PROXY_API_KEY and _LITELLM_PROXY_BASE_URL:
+    os.environ["LITELLM_PROXY_API_KEY"] = _LITELLM_PROXY_API_KEY
+    os.environ["LITELLM_PROXY_API_BASE"] = _LITELLM_PROXY_BASE_URL
+    litellm.use_litellm_proxy = True
```
pregnor, line 58: **Note**: to self, we need to centralize these settings and make it per agent overridable for convenient handling.

Edit: https://github.com/agntcy/coffeeAgntcy/issues/654

misi-bp (changes requested):

coffeeAGNTCY/coffee_agents/lungo/frontend/src/useApp.ts
pregnor: **Note**: LLM review comment

### 1. Clear conversation does not reset backend session (UX bug)

`handleClearConversation` only calls `chat.resetChatState()`; it does **not** rotate `patternChatSessionId`:

```389:395:coffeeAGNTCY/coffee_agents/lungo/frontend/src/useApp.ts
  const handleClearConversation = useCallback(() => {
    chat.resetChatState()
    streaming.resetGroup()
    streaming.resetRecruiter()
    lastDiscoveryKeyRef.current = null
    // eslint-disable-next-line react-hooks/exhaustive-deps -- stable reset fns only
  }, [chat.resetChatState, streaming.resetGroup, streaming.resetRecruiter])
```

In `pattern_doc` mode the UI clears, but the server keeps the ADK session for the same `session_id`. The next message continues prior context. **Fix:** mint a new `patternChatSessionId` on clear (and optionally call `usePatternChatAPI().cancel()` for in-flight streams).

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/Sidebar.tsx
pregnor, line 1: **Note**: LLM review comment

### 2. Supervisor / Peer Group / Recruiter stubs are not reachable from the UI

Reference Library entries come from **placeholder workflow patterns** (`use_case/scenario === "---"`), e.g. `"Feedback Loop"`, not architectural pattern names:

```180:214:coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/sidebar.utils.ts
const buildReferencePatternNames = (
  placeholders: readonly WorkflowSummary[],
  order: Map<string, number>,
): string[] => {
  const byPattern = new Map<string, number>()
  for (const row of placeholders) {
    // ...
    byPattern.set(row.pattern, idx)
  }
  return [...byPattern.keys()].sort(...)
}
```

The new `supervisor.md`, `peer_group.md`, and `recruiter.md` stubs serve `POST /patterns/Supervisor/chat`, but the sidebar never sends those names today. Either:

- add Reference Library entries for those three patterns, or  
- document that stubs are API-only until a follow-up lands.

Worth confirming against epic #588 intent before release.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/docs/workflows/peer_group.md
pregnor, line 1: **Required**: If we want to include the already implemented patterns in the reference library, we need to add them on the backend side to the starting topologies as unimplemented patterns - and I'm not sure how it will behave, because when I hacked the reference library into the system during end of Heartbeat, the intuition was that we have the set of patterns as full == implemented && unimplemented, not as full == unimplemented && full >= implemented.

Because of this, currently the implemented patterns, such as peer_group, recruiter & supervisor do not show up in the pattern reference library.

For that to happen we need to add them - in alphabetical order - to the rest of the reference library patterns at `coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/starting_workflows.json` with `use_case` & `scenario` set to "---", because that's how we mark unimplemented patterns, forcing them into the reference library.
Edit: also I'm not sure if it won't break anything, because at some places in FE we might be dependent on a unique pattern name - hence the original full == implemented+unimplemented understanding.

This is because we started from only patterns (implemented) and we hacked the unimplemented patterns (reference library) into the system in a short time period.
We could revise it, but we need more time for that.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Chat/ChatArea.tsx
pregnor, line 1: **Recommended**: Not sure if this is the right place for this, but I have noticed the chat response area is not centered properly, this should be looked at, but not a deal breaker.

Image

pregnor (changes requested):
Backend, schema: LGTM aside from the comments.

Frontend: I will leave this part to @misi-bp because he knows the new architecture/coding practices.

I tested it locally it looked ok (without the topology diagrams from the other PR), aside from the 3 missing implemented pattern reference and the chat area centering that I commented separately.

The change request is only because of the implemented references to get on the same page regarding expectations before merging.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/docs/workflows/peer_group.md
pregnor: **Note**: LLM review comment

### 2. Supervisor / Peer Group / Recruiter stubs are not reachable from the UI

Reference Library entries come from **placeholder workflow patterns** (`use_case/scenario === "---"`), e.g. `"Feedback Loop"`, not architectural pattern names:

```180:214:coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/sidebar.utils.ts
const buildReferencePatternNames = (
  placeholders: readonly WorkflowSummary[],
  order: Map<string, number>,
): string[] => {
  const byPattern = new Map<string, number>()
  for (const row of placeholders) {
    // ...
    byPattern.set(row.pattern, idx)
  }
  return [...byPattern.keys()].sort(...)
}
```

The new `supervisor.md`, `peer_group.md`, and `recruiter.md` stubs serve `POST /patterns/Supervisor/chat`, but the sidebar never sends those names today. Either:

- add Reference Library entries for those three patterns, or  
- document that stubs are API-only until a follow-up lands.

Worth confirming against epic #588 intent before release.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/Sidebar.tsx
pregnor: Resolving this as this is a duplicate to the manual test, I just forgot to remove this before sending the review.

Edit: added this to the other comment.

coffeeAGNTCY/coffee_agents/lungo/frontend/eslint.config.mjs
codyhartsook: Agreed. Removed the `ChatArea.tsx` exemption. The file is 369 lines so it's under the 400 limit and lint stays clean without it. Thanks for catching it.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/agenticWorkflowsApi.ts
codyhartsook: Left as `fetch` for now to stay consistent with `fetchWorkflowSummaries` in the same file. A clean axios migration of this module is better as its own sweep, noted for next time. The pattern-chat hook stays on `fetch` regardless, since it reads an NDJSON stream.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/useApp.ts
codyhartsook: Done. `CanvasMode` is now an exported `enum` and every comparison uses `CanvasMode.PATTERN_DOC` / `.WORKFLOW`.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/types/patternDoc.ts
codyhartsook: Good call. Made `CanvasMode` an exported `enum` (`WORKFLOW = "workflow"`, `PATTERN_DOC = "pattern_doc"`) and switched all comparisons to the enum members.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/utils/agenticWorkflowsApi.test.ts
codyhartsook: Confirmed. Only 404 maps to `WorkflowDocumentationNotFoundError`; 401/5xx and the rest fall through to a generic `Error` surfaced in the UI. Added an explicit 401 test to lock it in.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/hooks/useAppChatState.ts
codyhartsook: Collapsed into a single `if` with an `||`.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/PatternDoc/MermaidBlock.tsx
codyhartsook: Fixed. Colors now come from the OUK theme via `useTheme()` (`palette.primary.main`, `text.primary`, `action.hover`, etc.) and follow light/dark automatically. Also bumped edges to `primary.main` for better light-mode contrast.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/PatternDoc/PatternDocMarkdown.tsx
codyhartsook: Done, swapped to OUK `Box`. Audited the rest of `PatternDoc/*`; the only remaining raw elements are the mermaid SVG render island, which has to stay raw.

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/components/schemas.yaml
codyhartsook: Adopted `session://<uuid>`. Added the regex to `schemas.yaml` and a matching `pattern` on `PatternChatRequest` in `dtos.py`, FE now mints `session://${uuid()}`, and tests are updated including a malformed to 422 case. Verified the minted value matches the regex.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/pattern_chat.py
codyhartsook: Ack, left as-is and tracked in #654.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/useApp.ts
codyhartsook: Good catch. `handleClearConversation` now mints a fresh `session_id` in `pattern_doc` mode, so clearing starts a genuinely fresh backend session.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/Sidebar.tsx
codyhartsook: Confirmed and intentional for this PR. The Reference Library lists only unimplemented placeholders, so the Supervisor / Peer Group / Recruiter pattern docs are API-only (served via `POST /patterns/{name}/chat`) for now, and I documented that status in each stub. Surfacing implemented patterns in the Reference Library (adding them to `starting_workflows.json` as `---`) is the larger change with the unique-pattern-name risk you flagged, so I'd like to do that as a follow-up issue rather than here. Does that work for clearing the change request?

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/docs/workflows/peer_group.md
codyhartsook: Confirmed and intentional for this PR. The Reference Library lists only unimplemented placeholders, so the Supervisor / Peer Group / Recruiter pattern docs are API-only (served via `POST /patterns/{name}/chat`) for now, and I documented that status in each stub. Surfacing implemented patterns in the Reference Library (adding them to `starting_workflows.json` as `---`) is the larger change with the unique-pattern-name risk you flagged, so I'd like to do that as a follow-up issue rather than here. Does that work for clearing the change request?

codyhartsook: Same as the Reference Library thread above: documented these as API-only for now, with a proposed follow-up to surface implemented patterns in the sidebar.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Chat/ChatArea.tsx
codyhartsook: Fixed. Added `scrollbarGutter: "stable both-edges"` and aligned the thread and composer centering (both `maxWidth: 1100`, `mx: auto`). Should be centered now, let me know if it still looks off.

codyhartsook:
Thanks for the thorough reviews, and no worries on the draft timing. Addressed all the FE and schema comments: `CanvasMode` enum, eslint exemption removed, mermaid colors pulled from the theme, OUK `Box`, a 401 test, the `session://` id format, the clear-conversation session-reset bug, and chat centering. Rebased onto `main` so the mermaid diagrams from #652 are included. Replied inline on each thread.

The one open product question is the implemented-pattern references: I documented those stubs as API-only for now and am proposing a follow-up issue to surface implemented patterns in the Reference Library (rather than taking on the unique-pattern-name risk in this PR). Flagging that for your call since it was the main blocker.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/Sidebar.tsx
pregnor: Sure thing, fine by me.

coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/docs/workflows/peer_group.md
pregnor: Sounds good, thanks.

pregnor: Tracker: https://github.com/agntcy/coffeeAgntcy/issues/660

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/Sidebar/Sidebar.tsx
pregnor: Tracker: https://github.com/agntcy/coffeeAgntcy/issues/660

pregnor:
> Thanks for the thorough reviews, and no worries on the draft timing. Addressed all the FE and schema comments: `CanvasMode` enum, eslint exemption removed, mermaid colors pulled from the theme, OUK `Box`, a 401 test, the `session://` id format, the clear-conversation session-reset bug, and chat centering. Rebased onto `main` so the mermaid diagrams from #652 are included. Replied inline on each thread.
> 
> The one open product question is the implemented-pattern references: I documented those stubs as API-only for now and am proposing a follow-up issue to surface implemented patterns in the Reference Library (rather than taking on the unique-pattern-name risk in this PR). Flagging that for your call since it was the main blocker.

Tracker: https://github.com/agntcy/coffeeAgntcy/issues/660

pregnor (approved):
Fixes confirmed, mergable by me, wait for @misi-bp 's approval.

misi-bp (approved):
lgtm. thank you for the fixes

Issue #653, opened 2026-06-19
[Feature]: Implement UI improvements after review on 6/18


jparello:
### Problem Statement

See screenshot for areas 1,2,3,4 

- Use same font and font size in areas 1,2,4. This nets out to Area 4 font being bigger 
- Do not use bold font for selected items in area 2 (selected items)
- In area 2 only put border around the select item. Use the same color as the mouse over border area (hit should be lighter and in this shot only the "A2A Slim" would have the border around it/
- increase the indentations for the section text. For example "Conversation: Purchasing" should be at least 4 characters indented from "Supervisor", so under the P in supervisor.
- Nice to have move the collapse arrow to be on the left hand side of the border. So at the beginning of the line not the end
- In area 5 the prompt space. Remove the yellow borders and only use a yellow border when the prompt entry space has been filled in. So remove the yellow around the "Suggested Prompts" button, the collapse icon and the trash icon.  Highlight the prompt entry as yellow only when the text was copied to it. Remove it after it was sent.
- In area 5 increase the space for the prompt area on the bottom to twice its size.
- Nice to have.  Remove the collapse button and make the section expandable like the side panel
- Colors 
   - The color of the components in section 3 in dark and light mode should be the same color as the highlight color for the selected item in the left hand menu (2)
   - The background color for the prompt and reply transcript (section 4 and 5 ) should be the same color as the selected color in the left hand menu (2)
   - Remove the background box around the last reply  in the reply transcript. In the screenshot it's the order line with the grafana icon

<img width="1784" height="1192" alt="Image" src="https://github.com/user-attachments/assets/899a4b80-21c1-486e-a332-c66ba6516f86" />

### Proposed Solution

see above

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

misi-bp:
@jparello 
Please check my answer in this PR: https://github.com/agntcy/coffeeAgntcy/pull/662
1. If you like to implement the extendable chat area feature, let's talk about it, or make the behavior clear for me
2. Please check if there's still any need to change colors

I'd like to close this ticket once both above are implemented or skipped.

misi-bp:
The resizable chat area is covered here: https://github.com/agntcy/coffeeAgntcy/pull/715

PR #662, opened 2026-06-23
feat: UI improvements from John

misi-bp:
# Description

Lungo frontend UI polish for the workflow sidebar, agent chat area, and workflow graph canvas. Aligns visual hierarchy, typography, and interaction patterns across these panels for a more consistent experience.

## Replies to John's task one by one:

General comment: Using a design system has many advantages, but now I only emphasise two of them: accessibility and user experience. When a design system is created, the color palettes and their application guarantees that a person with eye sight issues can make difference between colors - both in light mode and dark mode. As for the user experience, for the choice of the colors of the different states is neither a coincidence: for an interactive element (button, input field, toggle switch) the color should be more vivid, thus when they are in focus or just clicked, they have to have the proper color. So for both topics, it is true, that the advantage of the unification will be lost if you want me to customize everything, especially the colors of the elements and their different states. This is true to all the icon buttons (like GH documentation, canvas control buttons): without a background, they look like simple icons without any interactive possibilities, they can’t be differentiated from the canvas elements without a background.

1. With removing the lighlighting background of the parent categories of a selected workflow in the sidebar, if we close the parent category, we can’t see which workflow is open on the canvas and where it is. Definitely not user friendly, I don’t recommend this, but implemented.
2. The color of the highlighting background around the selected items has not been changed, since it is a UX pattern that the hover status of an element should be less interesting, then an active status of that element. If you insist on changing it, please be more specific which color should be used for each state (hovered, just clicked, selected with the use of tab key).
3. Indentation in the sidebar is increased as you wanted it.
4. Collapse button is moved to the left side of the chat area header.
5. Yellow background is removed around the collapse and delete icon buttons, changed the ‘suggested prompts’ button to be a primary variant instead of ‘outlined’. Yellow background appears around text input field when there’s a content (when the prompts is sent, both the content gets removed and the zellow border with it.)
6. Last message (prompt reply) background highlighting is removed.
7. chatarea background is fixed to use of the highlighted selected menu item.
8. The color of the nodes in the canvas in dark and light mode have the same color as the highlight color for the selected item in the left hand menu
9. The text input for the user prompt takes up the available width

By areas:

### Sidebar
- Remove parent-category selection highlighting — only the selected workflow row is emphasized; pattern and scenario dropdown toggles no longer receive a `selected` state when a descendant workflow is active.
- Unify font weight for selected items (remove bold `fontWeight: 600` on selected rows).
- Increase nested dropdown indentation (`theme.spacing(1)` → `theme.spacing(3)`).
- Remove unused `patternContainsSelectedWorkflow` / `scenarioContainsSelectedWorkflow` helpers.

### Chat area
- Use `theme.palette.action.selected` as the chat panel background (replaces `getChatAreaBackgroundColor` / `background.default`); remove `chatAreaBackground.ts`.
- Standardize chat typography from `body2` to `body1` across messages, feeds, status lines, prompts dropdown items, and error text.
- **Chat header:** move minimize/collapse control to the left and clear conversation to the right; share horizontal padding with the message thread and composer; simplify icon button styling (remove outlined border treatment).
- **Composer:** full-width layout (remove `maxWidth: 880`); suggested-prompts trigger uses `primary` variant; inactive state handled via opacity in the dropdown component; input field shows a tertiary border when text is entered.
- Remove highlight styling from the final agent reply message.

### Workflow graph
- Simplify graph node resting background to `theme.palette.action.selected` for all nodes (drops per-state light/dark branching).
- Match node handle fill color to the same surface token.

### Tooling
- Declare `node: ">=24.0.0"` in `package-lock.json` `engines` (alongside existing `npm` constraint).

## Issue Link

Relates to #653

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] Existing issues have been referenced (where applicable)


pregnor (dismissed):
Code-wise LGTM, the visuals I was fine with those before and seeing you two having such strong opinions, I just want to move forward.

codyhartsook (dismissed):
lgtm

pregnor (approved):

mihaialexandrescu (approved):

PR #664, opened 2026-06-23
feat(lungo): A2A transport rail on agent nodes

### LINKED ISSUE #589: [Feat] Fiber: A2A transport rail on agent nodes
Parent: #592

## Summary

Highlight that Lungo agents communicate over **multiple A2A transports** depending on the agentic pattern in use (point-to-point `slimrpc`, group/pub-sub `slim`, broadcast `nats`, HTTP `jsonrpc`). Each agent's ReactFlow node grows a **Transport Rail** down its left edge: coloured vertical segments — one per `AgentInterface` in `additional_interfaces` — that on node hover extend a label row to the left showing the transport name and an active badge.

Data source is the existing **OASF endpoint** on backend agent supervisors (keyed by agent slug). The OASF record composes the A2A card, which carries `additional_interfaces` with each interface's `transport` and `url`. No new backend work.

**Scope:** Frontend only.

## Goals

- Every agent node in the workflow graph view exposes the A2A transports available on that agent via a vertical Transport Rail on the node's left edge.
- One coloured segment per `AgentInterface` in the agent card's `additional_interfaces`.
- On node hover, each segment extends to the left with a label row showing the transport name and an active badge.
- The feature makes the multi-transport story visible: e.g. `slimrpc`, `slim`, `nats`, `jsonrpc` on the same agent.

### Data source

- Existing OASF endpoint on backend agent supervisors, keyed by agent slug. Returns an OASF record that composes the A2A card.
- The A2A card carries `additional_interfaces`: a list of `AgentInterface` objects with at least `transport` (e.g. `slimrpc`, `slim`, `nats`, `jsonrpc`) and `url`.
- Slug is already present on agent nodes in the topology payload from #430.

### UI: Transport Rail

- Custom ReactFlow agent node gains a Transport Rail along the **left edge**.
- One **coloured vertical segment** per `AgentInterface`. Segment colour is per-transport (consistent across all nodes so the user learns the mapping at a glance).
- On node hover, each segment extends to the **left** with a label row showing:
  - transport name (e.g. `slimrpc`)
  - an "active" badge indicating which transport this agent is currently selected for in the running workflow (when that information is available)
- Default (non-hover) state shows segments only — no labels — to keep the node compact.

### "Active" badge

- "Active" reflects the transport currently selected by the A2A client factory for that agent in the current workflow instance.
- If the active transport isn't reliably knowable from existing state in fiber, render the rail without the active badge and treat the badge as a follow-up. Diagnose during implementation; don't block the rail itself.

### Fetch strategy

- One OASF lookup per distinct slug; cache for the lifetime of the workflow instance view.
- Missing/unreachable OASF records: rail simply doesn't render — no error state.

## Child stories / sub-items

1. **OASF fetch + cache in the FE** — call the existing OASF endpoint by agent slug; cache per workflow instance view; handle missing records gracefully.
2. **Transport colour mapping** — define a stable colour per known A2A transport (`slimrpc`, `slim`, `nats`, `jsonrpc`, plus a default for unknowns). Single source of truth in the FE.
3. **Transport Rail on the agent node** — render one coloured segment per interface along the left edge of the custom ReactFlow node.
4. **Hover label rows** — on node hover, extend each segment with a label row to the left showing transport name (and active badge when available).
5. **Active transport indication** — wire the "active" badge to whatever state already indicates the client factory's current selection for the agent; if not available, ship without it and track follow-up.

## Acceptance criteria

- Each agent node renders a vertical rail on its left edge with one segment per `AgentInterface` in its OASF-derived A2A card.
- Segment colours are consistent per-transport across the graph.
- Hovering a node extends each segment with a label row showing the transport name.
- Missing or unreachable OASF records do not break the graph.

codyhartsook:
Closes #589

## Summary

- OASF records for all A2A agents updated with the latest A2A card data via `dirctl import` from the A2A AgentCard, with an LLM enricher pass and manual editing. This was necessary to include the `additional_interfaces` transport entries that power the new Transport Rail info bars.
- Adds a **Transport Rail** along the left edge of each ReactFlow agent node, rendering one coloured vertical segment per `AgentInterface` in the agent's OASF-derived A2A card (`additional_interfaces`).
- On node hover, each segment extends left with a label row showing the transport name (e.g. `slimrpc`, `slim`, `nats`, `jsonrpc`).
- Adds `useNodeTransportInterfaces` hook — fetches and caches OASF data per agent slug for the lifetime of the workflow instance view; missing/unreachable records silently skip the rail.
- Adds `transportMeta.ts` — single source of truth for per-transport colour mapping.

https://github.com/user-attachments/assets/e13c83b6-350a-4a27-8d81-383397d24127

## Changed files

| Area | Files |
|---|---|
| Transport Rail UI | `CustomNode.tsx` (+221 lines), `transportMeta.ts` (new), `types.ts` |
| OASF fetch/cache hook | `useNodeTransportInterfaces.ts` (new), `useMainArea.ts` |
| OASF data | 7 agent card JSON files updated with `additional_interfaces` |
| Tests | `transportMeta.test.ts` (new, 105 lines) |

## Test plan

Tested end-to-end with docker-compose.

- [ ] Run `transportMeta.test.ts` — all transport colour and metadata assertions pass
- [ ] Open the workflow graph view — each agent node shows coloured segments on its left edge
- [ ] Hover an agent node — label rows with transport names extend to the left
- [ ] Agent with missing OASF record — rail does not render, graph remains stable
- [ ] Verify colour consistency: same transport produces the same colour across all nodes


coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/MainArea/Graph/Elements/CustomNode.tsx
```diff
@@ -47,8 +50,193 @@ interface CustomNodeProps {
...
+  })
+
+  return `linear-gradient(to bottom, ${stops.join(", ")})`
+}
+
+const TransportRail: React.FC<{
+  transports: AgentTransport[]
+  expanded: boolean
+  activeTransport?: string
+}> = ({ transports, expanded, activeTransport }) => {
```
delthazor: You may extract this to a separate file, to avoid the 'File too long' warning.

I mean, I don't think it is necessary, this looks like a coherent unit, and the 400 lines limit is rather small IMHO, and it is just a warning, not an error, but if that is the limit we would like to impose on frontend files, then this could be extracted, with relatively small effort.

(Although the helper function may need some extra work to have them working both inside TransportRail, and outside in the CustomNode...)

pregnor (dismissed):
LGTM, get an approval from @misi-bp on the FE pls.

coffeeAGNTCY/coffee_agents/lungo/frontend/src/components/MainArea/Graph/Elements/CustomNode.tsx
codyhartsook: Agree with the suggestion, created a separate component file for info rail component

delthazor (approved):
lgtm

pregnor (approved):
Still wait for @misi-bp 's approval

misi-bp (approved):
tested. lgtm

pregnor:
@codyhartsook I found 1 thing while smoke testing.

When I `docker compose up --build [--detach]` on this branch and the services come up, for me 3/3 times recruiter-supervisor dies with the following after uvicorn starts serving `recruiter-supervisor exited with code 137` after the 2 empty line logs.

If I manually restart the service then it comes up fine and serves fine from then on.

I tried the same on main but there it did not occur even with the SLIM 1.4 upgrade.

I could not figure out what on earth in this PR could affect that happening, so I'm leaning a bit towards it being local to my machine, but I would ask for an independent repro attempt.
If it's not constantly occurring elsewhere then this is something we can investigate separately.

Edit: I also checked docker resources, didn't seem to be the constrain here.

Edit: Ah, ok, nevermind, I can see a spike in memory just when recruiter-supervisor is becoming ready and then docker (even dashboard) cycles something, because I lose resource usage telemetry data, so it's most likely memory spikes, but I will try to repro again.

Edit: yeah, right before it getting killed - I guess - memory spikes from around 450MB to around 750MB+ and the overall consumption gets closer than 500MB to the limit, so it's probably that, but I will have to look deeper.

Edit: yes, confirmed OOMKilled, so must be a local issue, though the memory spike could be looked into, but probably not PR related.

codyhartsook:
> @codyhartsook I found 1 thing while smoke testing.
> 
> When I `docker compose up --build [--detach]` on this branch and the services come up, for me 3/3 times recruiter-supervisor dies with the following after uvicorn starts serving `recruiter-supervisor exited with code 137` after the 2 empty line logs.
> 
> If I manually restart the service then it comes up fine and serves fine from then on.
> 
> I tried the same on main but there it did not occur even with the SLIM 1.4 upgrade.
> 
> I could not figure out what on earth in this PR could affect that happening, so I'm leaning a bit towards it being local to my machine, but I would ask for an independent repro attempt. If it's not constantly occurring elsewhere then this is something we can investigate separately.
> 
> Edit: I also checked docker resources, didn't seem to be the constrain here.
> 
> Edit: Ah, ok, nevermind, I can see a spike in memory just when recruiter-supervisor is becoming ready and then docker (even dashboard) cycles something, because I lose resource usage telemetry data, so it's most likely memory spikes, but I will try to repro again.
> 
> Edit: yeah, right before it getting killed - I guess - memory spikes from around 450MB to around 750MB+ and the overall consumption gets closer than 500MB to the limit, so it's probably that, but I will have to look deeper.
> 
> Edit: yes, confirmed OOMKilled, so must be a local issue, though the memory spike could be looked into, but probably not PR related.

could not replicate locally

Issue #682, opened 2026-06-30
Refactor the full stack (session) context layer + ID handling


pregnor:
Currently it's a big mess and when we have sessions, conversations, agent memory, requests, tracing it becomes a living hell as highlighted by my work on #671.

I made up a mental model of a working system that would fit the current boundaries with necessary improvements and clear separation of responsibilities, but that requires a substantial refactor, so I made the LLM work out the details and stash it for later so I can go back and do the hotfix for Fiber and then move on to other burning priorities.

**Note**: I would use the URI scheme (`type://uuid`) everywhere, the LLM was just not good enough to derive it and it's not worth reburning all the tokens on or my time to manually fix it.

**Note**: Frontend session does not necessarily equal browser tab lifecycle in my head, but today that's closest to what I mean, if we had users and sessions beyond tab opening, session would mean that long thing, independent from browser mechanics, but this was how I managed to describe it to the LLM concisely.

**Note**: Don't forget thorough, but concise docs of it once implemented, both from newcomer and change perspective including a comparison with the old model.

---

# Lungo prompt context ID system — complete design & implementation reference

Single reference for the **full** ID refactor: all chat patterns, pattern-doc chat, hierarchy persistence, Grafana rename, graph `conversation_id`, OpenAPI, and **no** deprecated `session_id` on supervisor responses (pattern-doc migrates `session_id` → `conversation_id` with a short input alias only).

**Motivation:** [#671](https://github.com/agntcy/coffeeAgntcy/issues/671) Grafana links; recruiter multi-turn; one mental model for every prompt path.

---

## 1. Executive summary

Lungo uses **four explicit IDs** at four lifetimes. Every prompt-sending path shares the same request/response envelope. Grafana uses **`request_id`** only. Agent/graph memory uses **`conversation_id`**. Workflow animation uses **`workflow_instance_id`**. Tab scope uses **`frontend_session_id`**.

---

## 2. ID hierarchy (ground up)

```
frontend_session_id          ← one per browser tab
  └── workflow_instance_id   ← one per workflow canvas run
        └── conversation_id  ← one clearable chat thread (agent/graph memory)
              └── request_id   ← one per prompt (OTEL executionID → Grafana)
```

| ID | Minted by | Lifetime | Consumer |
|----|-----------|----------|----------|
| `frontend_session_id` | Frontend (`sessionStorage`) | Tab | Logging, hierarchy grouping, future analytics |
[...]

PR #683, opened 2026-06-30
fix(#671,lungo,fe):  fix UI response grafana link

### LINKED ISSUE #671: Post-prompt Grafana links are only showing up on Peer Group prompt (regression)
It is expected to be shown for all the prompts.

Also heartbeat+ release (0.1.2) it was still working for Supervisor-worker / batch (but not streaming).

<details>

<summary>0.1.2 Peer group screenshot (working)</summary>

Image

</details>

<details>

<summary>0.1.2 Supervisor batch screenshot (working)</summary>

Image

</details>

<details>

<summary>0.1.2 Supervisor streaming screenshot (missing)</summary>

Image

</details>

<details>

<summary>0.1.2 Recruiter screenshot (missing)</summary>

Image

</details>

<details>

<summary>0.2.0 Peer group screenshot (working)</summary>

Image

</details>

<details>

<summary>0.2.0 Supervisor batch screenshot (missing)</summary>

Image

</details>

<details>

<summary>0.2.0 Supervisor streaming screenshot (missing)</summary>

Image

</details>

<details>

<summary>0.2.0 Recruiter screenshot (missing)</summary>

Image

</details>

pregnor:
# Description

1. Fixed grafana buttons after prompt.
2. Added a frontend test for it.
3. Also had to add trace ID for recruiter to make it work with persistent session for multiple prompt conversations. Created #682 for long term solution.

<details>

<summary>Button fix video</summary>

https://github.com/user-attachments/assets/b63f5136-7d42-406d-adab-b754e2e4d72d

</details>

<details>

<summary>No regression video</summary>

https://github.com/user-attachments/assets/a144b556-9f91-4393-8e73-acc2b985879b

</details>

## Issue Link

Fixes #671 

## Type of Change

- [X] Bugfix

## Checklist

### DIFF
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
@@ -1,6 +1,8 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
+import config.logging_config  # noqa: F401 - runs setup on import; must be first
+
 import asyncio
 import json
 import logging
@@ -8,7 +10,6 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 from typing import Optional
 
-import config.logging_config  # noqa: F401 - runs setup on import; must be first
 import httpx
 import uvicorn
 from agents.supervisors.recruiter import shared
--- a/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
+++ b/coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
@@ -1,50 +1,51 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
-import config.logging_config  # noqa: F401 - runs setup on import; must be first
-
 import asyncio
 import json
 import logging
 from contextlib import asynccontextmanager
+from pathlib import Path
 from typing import Optional
-from uuid import uuid4
 
+import config.logging_config  # noqa: F401 - runs setup on import; must be first
 import httpx
 import uvicorn
-from dotenv import load_dotenv
-from fastapi import FastAPI, HTTPException, Request
-from fastapi.middleware.cors import CORSMiddleware
-from fastapi.responses import JSONResponse, StreamingResponse
-from pathlib import Path
-from pydantic import BaseModel
-
-from common.cors import get_cors_allowed_origins
-
 from agents.supervisors.recruiter import shared
 from agents.supervisors.recruiter.card import RECRUITER_SUPERVISOR_CARD
 from agents.supervisors.recruiter.recruiter_client import get_a2a_event_queue
 from agents.supervisors.recruiter.recruiter_service_card import (
     RECRUITER_AGENT_URL,
 )
-from common.streaming_capability import require_streaming_capability
-from config.config import LLM_MODEL, HOT_RELOAD_MODE, OTEL_SDK_DISABLED
 from agntcy_app_sdk.factory import AgntcyFactory
+from common.cors import get_cors_allowed_origins
+from common.streaming_capability import require_streaming_capability
+from config.config import HOT_RELOAD_MODE, LLM_MODEL, OTEL_SDK_DISABLED
+from dotenv import load_dotenv
+from fastapi import FastAPI, HTTPException, Request
+from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import JSONResponse, StreamingResponse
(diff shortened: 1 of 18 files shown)


coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
```diff
@@ -128,165 +133,221 @@ async def handle_stream_prompt(request: PromptRequest, req: Request):
...
-            },
-        )
+                                "trace_id": trace_id,
+                            }
+                        )
+                        + "\n"
+                    )
+
+            return StreamingResponse(
+                stream_generator(),
```
delthazor, line 345: I might be mistaken, but this seems to be inside the 'with session_start()' scope.
Is it intentional?

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
```diff
@@ -1,50 +1,51 @@
...
-
 import asyncio
 import json
 import logging
 from contextlib import asynccontextmanager
+from pathlib import Path
 from typing import Optional
-from uuid import uuid4
 
+import config.logging_config  # noqa: F401 - runs setup on import; must be first
```
delthazor: You sure this is ok to reorganize?
Comment seems to be contradicting the reorder.

delthazor (dismissed):
My comments are just nitpicks, they can be discarded.

Otherwise it is ok from me.

mihaialexandrescu (dismissed):

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
pregnor: It is intentional, I indented the entire try block out of simplicity and because I wasn't sure if I will need to touch that part.

But now it is known I won't have to use the trace_meta there, so I'm fixing this, it's fine either way.

pregnor: Wait no, I'm leaving it that way rather, because the `stream_generator()` call is there, so I want to make sure the trace is still in scope when that happens as we need the trace_meta.executionID during that - edit: and also the other trace logics.

I'm not sure whether Python would handle it fine if I released it after the definition, before the call site, but I feel like this is safer anyway.

pregnor: Technically what I would do - unless infeasible - is to move the stream_generator definition out of the endpoint definition, but we would need to instantiate and pass down the trace_meta in that case, so logically the current setup is the more correct setup (we are using the trace runtime, not definition-time, so I rather want the trace_meta in scope runtime).

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
pregnor: It has to be first among non-stdlib imports, because our other libraries and other 3rd party libraries might depend on some of the log setup set here - so the logging is not like random at the beginning before this library is loaded.

My linter orders stdlib stuff above and other 3rd party stuff below, so IMO this should be fine.

Also I quite literally tested every single prompt, so this should be fine docker-compose-wise.
I don't think this should matter in docker-compose vs k8s context.

Also you are the author, so you tell me whether this comment is still relevant and whether it causes anything:

https://github.com/agntcy/coffeeAgntcy/commit/1e0445718d36dc6423ae45d1b5f5660902ce5693

(also it might be worth checking whether there is a better solution to this, but as it is logging I wouldn't worry too much until log processors are set onto the logs, but we have bigger fish to fry on that front)

delthazor: It was indeed familiar :D

I think it is enough to be imported before non-stdlib imports, just the comment was not specific in that regard. You can leave it like this, we can revisit the logging mechanics later.

pregnor: Fixed in [9b953c4](https://github.com/agntcy/coffeeAgntcy/pull/683/commits/9b953c4122918c80a47bc964809adce5c3a8d640).

coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/recruiter/main.py
delthazor: I'm fine with current version, just wanted to make sure, it is intentionally this way.

mihaialexandrescu (approved):

delthazor (approved):
lgtm

pregnor:
I'm going to wait for @misi-bp 's review because of the frontend.

misi-bp (approved):

PR #696, opened 2026-07-07
refactor(corto,lungo,rectruiter): unify docker files

### LINKED ISSUE #695: Minimize duplicates of dockerfiles
The agents docker build processes are very similar and can be merged together into more common dockerfiles to clean up the repository a bit.

arpad-csepi:
# Description

This PR consolidates all Dockerfiles into a single shared `coffeeAGNTCY/docker/` directory.

Previously each agent project kept its own per-service Dockerfiles; now parameterized templates cover all images. Eliminated per-service Dockerfiles across corto and lungo.

This templates introduces  docker build args:
  - `UV_CACHE_ID` and `NPM_CACHE_ID` for docker build cache.
  - `APP_NAME` for build metadata.
  - `PROJECT_PATH` for locating the project root path
  - `AGENT_SCRIPT` for locating the entrypoint of the agent

Documentation moved into per-project READMEs under `coffeeAGNTCY/docker/`.

## Issue Link

Fixes #695 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [X] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist


delthazor (dismissed):
I think .github/workflows/readme would also need some updates after this, but otherwise seems solid.

arpad-csepi:
> I think .github/workflows/readme would also need some updates after this, but otherwise seems solid.

Created a separate PR for this where I updated the readme, feel free to check it out or even force push it with changes.
https://github.com/agntcy/coffeeAgntcy/pull/697

mihaialexandrescu (dismissed):

pregnor (changes requested):
Move the folder from `coffeeAgntcy/coffeeAGNTCY/docker` to `coffeeAgntcy/docker` please.

That directory is already redundant - it makes sense from python packaging perspective, but it still looks bad there - and docker doesn't need the redundant path.

Edit: update .github/workflows/docker-build-push.yaml as well with the new path.
Edit: also the readmes, better do a search & replace on `coffeeAGNTCY/docker`->`docker`

coffeeAGNTCY/docker/Dockerfile.recruiter
```diff
@@ -6,6 +6,14 @@ FROM astral/uv:0.11.16-python3.13-trixie-slim@sha256:d57fc364ed714127c162a01b316
 
 COPY --from=ghcr.io/agntcy/dir-ctl:v1.5.0 /dirctl /usr/local/bin/dirctl
```
pregnor: **Question**: If I see it right, this is the only difference compared to the other agent images, have you looked into whether we can make this a build argument? Are there no generic solutions where we would just feed the image, source location, target destination?

pregnor (commented):
Aside from the directory location it looks perfect, thank you very much.

pregnor (approved):

coffeeAGNTCY/docker/Dockerfile.recruiter
arpad-csepi: Made the `Dockerfile` multi-staged one, so with a target argument can be switched between corto/lungo and recruiter build. So the important note here is the agent-with-dirctl now also do extra (exposes port) so might not fit later well for other agents and needs further refinement.

mihaialexandrescu (approved):

Issue #595, opened 2026-05-27
External contributors should receive full integration test results


delthazor:
### Problem Statement

Currently, external contributors can not get a green CI result, because integration test secrets are not injected on their forks.

### Proposed Solution

For every Pull Request, a safe smoke test suite should be running without conditions.
After that, a code owner or similar trusted user could do a review and approve the PR for the full test suite. When approved this way, the CI should run a single time with the secrets injected, so the contributors can get a proper integration test result.

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

pregnor:
We should split the test suite into tests that don't need LLM API keys (working with mocked input/output to test integration) and ones that do.

We should set up the testing so CI/CD runs the keyless suite while we should automate running the keyed tests locally for external contributor PRs and releases.

pregnor:
This was done in #707 as much as we could for now.
#734 will fix an issue that presumably was caused by #707 but I close this one in the meantime.

PR #707, opened 2026-07-21
Split CI tests into separate tiers

delthazor:
# Description

Replace the monolithic integration job with multiple tiers so distinct situations like fork PRs can use a different set based on specific needs.

The goal of this PR is to prepare the test suite for later configuration for specific environments and use cases, like for PR runs.

Important: tests requiring LLM secrets are left out from CI suite, due to security concerns. The test are still part of the codebase, and can be ran locally by manually calling them.

UPDATE: Renamed issue, as the basis of the split was changed from markers to directories. Description changed accordingly.

## Issue Link
Closely relates to (but not fully fixes): #595 

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist


.github/workflows/README.md
```diff
@@ -103,13 +157,15 @@ jobs:
...
+Use `test_mode: llm-only` only when you intentionally want to skip fast and docker-only integration tests.
+
 ## fe-ci
 
 Runs frontend checks (TypeScript typecheck, ESLint, Prettier via `npm run check`) for the Lungo frontend. Only triggers when files under `coffeeAGNTCY/coffee_agents/lungo/frontend/` change.
 
 ## version-override-test
 
-Demonstrates how to pin or constrain dependencies and override container images when calling the reusable test workflow.
+Demonstrates how to pin or constrain dependencies and override container images when calling the reusable test workflow. Uses the default `test_mode: all` so fast, docker, and LLM tiers all run against the overridden images.
```
pregnor: **Question**: If a test mode all exists and is the default shouldn't we also document it?

pregnor (commented):
I'm not on board with the marker approach, I would favor a suite or test package level separation of the tests over these markers.

I'm also against the dynamic marker application code like in `pytest_collection_modifyitems`, because it makes it difficult to read the code effectively.

Also let's eliminate LLM key based tests from the CI and do those manually, that simplifies things in the CI workflows as well.

.github/workflows/README.md
delthazor: The question was valid, but not applicable any more. Structure was changed so test_mode no longer exists.

coffeeAGNTCY/coffee_agents/corto/tests/README.md
```diff
@@ -2,55 +2,73 @@
...
 
-- Session / infra orchestration fixtures & agent/client fixtures: [`conftest.py`](coffeeAGNTCY/coffee_agents/corto/tests/integration/conftest.py:1)
-- Docker Compose lifecycle helpers (bring up transport and observability components): [`docker_helpers.py`](coffeeAGNTCY/coffee_agents/corto/tests/integration/docker_helpers.py:1)
-- Lightweight subprocess runner used for agent processes: [`process_helper.py`](coffeeAGNTCY/coffee_agents/corto/tests/integration/process_helper.py:13)
-- Sommelier (flavor profile) integration test: [`test_sommelier.py`](coffeeAGNTCY/coffee_agents/corto/tests/integration/test_sommelier.py:1)
+| Directory | Purpose | CI |
+|-----------|---------|-----|
+| `tests/unit/` | Mocks only | Yes |
+| `tests/integration/` | Docker-compose session; no LLM | Yes |
+| `tests/integration_llm/` | Docker + LLM credentials | No (local manual) |
```
pregnor: **Question**: This differs from .github/workflows/README.md even though structurally live should also be present here, because it's similar architecture, is it because we have no tests in that category or something else?

.github/workflows/README.md
```diff
@@ -70,21 +71,79 @@ Reusable workflow called by `helm-push.yaml` for each matrix entry. Accepts:
...
+| `tests/integration/` | Docker-compose session; no LLM | Yes |
+| `tests/integration_llm/` | Docker + LLM credentials | No (local manual) |
+
+### Per-project CI pytest paths
+
+| Project | `test_paths` |
+|---------|--------------|
+| corto | `tests/unit tests/integration` |
+| lungo | `tests/unit tests/live tests/integration` |
+| recruiter | `tests/unit tests/integration` |
```
pregnor: **Question**: Shouldn't these be analogous with the naming above?

Also shouldn't we use single word folder names instead?

.github/workflows/README.md
```diff
@@ -70,21 +71,79 @@ Reusable workflow called by `helm-push.yaml` for each matrix entry. Accepts:
...
+# CI-equivalent (no secrets)
+cd coffeeAGNTCY/coffee_agents/corto && uv run pytest tests/unit tests/integration -q
+cd coffeeAGNTCY/coffee_agents/lungo && uv run pytest tests/unit tests/live tests/integration -q
+cd coffeeAGNTCY/coffee_agents/recruiter && uv run pytest tests/unit tests/integration -q
+
+# LLM (local only, needs .env)
+cd coffeeAGNTCY/coffee_agents/lungo && uv run pytest tests/integration_llm -q
+```
+
+Do not run `pytest tests/` as a single full-suite invocation when both `integration/` and `integration_llm/` exist — session fixtures may load twice via `pytest_plugins`.
```
pregnor: **Question**: why is this a requirement?
I thought it's going to be disjunctive sets that together gives a whole.

.github/workflows/README.md
```diff
@@ -70,21 +71,79 @@ Reusable workflow called by `helm-push.yaml` for each matrix entry. Accepts:
...
 
-Reusable job that runs `pytest` via `uv` for a single project directory. Exposes inputs for dependency and Docker image overrides.
+### Branch protection
+
+| Remove (legacy) | Add (required on PRs) |
+|-----------------|----------------------|
+| `integration-tests-*` | `tests / *` |
+| `fast / *`, `integration / *` | `tests / *` |
```
pregnor: **Question**: What is this for?

.github/workflows/README.md
```diff
@@ -75,42 +75,44 @@ Reusable workflow called by `helm-push.yaml` for each matrix entry. Accepts:
...
 
 ### Test directories
 
 | Directory | Meaning | CI |
 |-----------|---------|-----|
 | `tests/unit/` | Mocks only | Yes |
-| `tests/live/` | Subprocess uvicorn/A2A (lungo only) | Yes |
-| `tests/integration/` | Docker-compose session; no LLM | Yes |
-| `tests/integration_llm/` | Docker + LLM credentials | No (local manual) |
+| `tests/integration/general/` | Docker-compose session; no live webserver; no LLM | Yes |
```
pregnor: **Recommended**: (should) `tests/integration/general/*`->`tests/integration/*` (no need for redundant general stuff).

Educate me if there is any opposing reason.

This needs an update in other files as well.

.github/workflows/README.md
```diff
@@ -75,42 +75,44 @@ Reusable workflow called by `helm-push.yaml` for each matrix entry. Accepts:
...
 ### Per-project CI pytest paths
 
 | Project | `test_paths` |
 |---------|--------------|
-| corto | `tests/unit tests/integration` |
-| lungo | `tests/unit tests/live tests/integration` |
-| recruiter | `tests/unit tests/integration` |
+| corto | `tests/unit tests/integration/general` |
+| lungo | `tests/unit tests/integration/general tests/integration/live` |
+| recruiter | `tests/unit tests/integration/general tests/integration/live` |
```
pregnor: **Recommended**: (should) `tests/unit tests`->`tests/unit/*` let's skip spaces and also the word `tests` are redundant.
If we kept `general_tests` I would advise `general` as well, but see the above command why to fold that up.

Educate me if there is an opposing reason.

This needs an update in other files as well.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_transport_config.py
```diff
@@ -26,6 +23,10 @@ def build_a2a_client_config(
     Build ClientConfig for A2AClientFactory using SLIM_SERVER, NATS_SERVER (if enabled),
     and SLIM_SHARED_SECRET from the environment.
     """
+    # Read transport endpoints at call time so test fixtures can monkeypatch env and
+    # reload supervisor modules without stale import-time values from .env (e.g. nats:4222).
+    from config.config import NATS_SERVER, SLIM_SERVER
+
```
pregnor: **Question**: Why are we moving this here? I guess it's lazy loading of env, but we should reload in that case IMO.
I have a note to fix these. it can fly for now, just noting.

coffeeAGNTCY/coffee_agents/lungo/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "langchain-litellm==0.3.5",
     "llama-index-llms-litellm==0.6.3",
     "llama-index-core==0.14.9",
+    "griffe<2",  # banks/llama-index import griffe 1.x API; 2.x is a meta-package without it
```
pregnor, line 32: **Question**: Did you confirm this manually?

coffeeAGNTCY/coffee_agents/lungo/common/a2a_transport_config.py
pregnor: Proposal: https://github.com/agntcy/coffeeAgntcy/pull/727

coffeeAGNTCY/coffee_agents/corto/tests/README.md
delthazor: For Corto, this category is empty, and that's why it is undocumented. (also, structure changed since)

.github/workflows/README.md
delthazor: yes we should, and the new structure accommodates that.

.github/workflows/README.md
delthazor: I agree that this is ambiguous. The intent here was that the integration tests are run by CI, so manual run of the integration_llm tests does not need to duplicate that.

Nevertheless, this changed again since.

.github/workflows/README.md
delthazor: I think that is a leftover marking that documents what needs changing in the organizational admin repo. This needs to be removed from here.

.github/workflows/README.md
delthazor: Since live and llm got a dedicated folder, it felt right to put all the other tests into a separate folder as well, as using the integration root for both as a subfolder container and a test package in itself.

I can be convinced otherwise, if you insist on that simplified arrangement.

.github/workflows/README.md
delthazor: There is no opposing reason, you are right on this one, although I think this is just a documentation error, the actual folder structure does not contain the redundant "test" phrase and the spaces.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_transport_config.py
delthazor: I agree with the proposal. I will rebase the solution on that, and update the corresponding parts.

coffeeAGNTCY/coffee_agents/lungo/pyproject.toml
delthazor: Yes, in the sense that running 'uv run pytest' with a newer griffe version produced missing package errors, and I had to downgrade my version to get it running, after I verified it as a transitive dependency that I can not remove.

But I did not manually checked the actual contents of the package in either versions, and did not check the release notes on them, I relied on Cursor's reasoning on why it works this way.

.github/workflows/README.md
pregnor: I would opt for defaults being in unit/integration/e2e and only put special one in subpackages, because that's the common setup, I feel like general or similar naming doesn't add any information, but now you need to use it.

Check whether we can run only the parent package (integration) easily without including the subpackages, if we can I would go for folding up.

coffeeAGNTCY/coffee_agents/lungo/pyproject.toml
pregnor: That is enough, thanks.

pregnor: I only meant that it's a real dependency, not a hallucination.

.github/workflows/README.md
delthazor: I checked and it seems, with `testpaths = ["tests"]` in pyproject.toml, pytest tests/integration recurses into all subdirs, so it would also collect live/ and llm/.

The fold is feasible only if CI uses explicit exclusion, e.g.:

```
pytest tests/unit tests/integration/live \
  tests/integration --ignore=tests/integration/live --ignore=tests/integration/llm
```
Or list root-level test modules explicitly (fragile as files are added).

Although the 'general' folder could be renamed to something more descriptive, like 'primitives' or 'internal', or as a contrary to the already used 'live', this could be called 'offline'. With properly descriptive name I think these basic tests could remain in a separate directory.

.github/workflows/README.md
delthazor: Documentation files are updated.

coffeeAGNTCY/coffee_agents/lungo/common/a2a_transport_config.py
delthazor: I rebased the solution, it should include your merged proposal now.

.github/workflows/README.md
pregnor: I would go for being explicit on the CI then

.github/workflows/README.md
delthazor: Removed

.github/workflows/README.md
delthazor: Folded up then, CI call made explicit.

pregnor (approved):

.github/workflows/test.yaml
```diff
@@ -90,138 +96,88 @@ on:
...
       docker_overrides: ${{ inputs.docker_overrides }}
-    secrets: inherit
 
-  integration-tests-recruiter:
+  tests-lungo-path-skip:
+    name: tests / lungo
     needs: subprojects-to-test
-    if: ${{ needs.subprojects-to-test.outputs.test_recruiter == 'true' }}
+    if: |
+      needs.subprojects-to-test.outputs.test_lungo != 'true' &&
```
mihaialexandrescu, line 152: Can you explain please what this this condition is useful for? We are isolating tagging ?
I'm asking because per the code in `test-subprojects-reusable.yaml` (moved from here, same otherwise) a "push" event always sets `test_lungo` to `true`.

mihaialexandrescu (approved):

.github/workflows/test.yaml
delthazor: The `-path-skip` jobs aren't for tag isolation — they're stubs so required branch-protection checks named tests / lungo (etc.) still report success when PR path filtering skips that subproject. Without them, a PR that only touches corto would never produce a tests / lungo check and could block merge. The path-skip job uses the same display name (tests / lungo) and exits immediately with a log message, so the required check is satisfied without running pytest. The real and skip jobs are mutually exclusive.

You're right that on push, test-subprojects-reusable.yaml always sets all flags to true, so the path-skip branch never fires on push today — the `|| github.event_name == 'push'` guard is redundant given current logic. We could drop the push arm for clarity, or change push to use the same path filtering as PRs if we want to avoid running unchanged projects on main.

mihaialexandrescu: I think, in general, we want to run the tests for merges on `main` even if we don't have code changes for that subproject. It's just for PRs that we want to skip running unchanged subprojects.

Thanks for the explanation.

PR #726, opened 2026-07-29
694/open api spec  status codes

### LINKED ISSUE #694: Update OpenAPI spec with all status codes from the backend code
This issue stems from this PR thread: https://github.com/agntcy/coffeeAgntcy/pull/680#discussion_r3496856478.

The goal is to have all the relevant codes listed in the OpenAPI spec (not just a `200`). We are only targetting this for Lungo.

In order to maintain consistency over time, we may need to also update the Skill that PR creates (`.agents/skills/api-documentation-lungo/`) so that it can suggest updates to the OpenAPI specs too based on the code (not just write to the documentation file the status codes discovered only in the backed code) since it reasons and discovers them anyhow for the documentation.

---

### Example:

Currently, the OpenAPI spec only lists very few of the status codes that are possible to come out of the backend code for our workflow-instance API endpoints.
For example:
https://github.com/agntcy/coffeeAgntcy/blob/9c36ad3257318d37a0155a8cdb175b0e8212f02b/coffeeAGNTCY/coffee_agents/lungo/schema/openapi/paths/agentic-workflows.yaml#L148-L158

However, the backend code can return a number of 4xx or 5xx status codes (different for each API endpoint) depending on what it does and what errors it encounters.
For example, we can get a `500` from this bit of code: (`ValueError` in the seed builder → caught in the handler → HTTP `500` "Workflow catalog inconsistency".)
https://github.com/agntcy/coffeeAgntcy/blob/9c36ad3257318d37a0155a8cdb175b0e8212f02b/coffeeAGNTCY/coffee_agents/lungo/api/agentic_workflows/instance_lifecycle.py#L41-L46

Overall, that example `POST /agentic-workflows/{workflow_name}/` endpoint could return (according to current code):
- `200` — instance created; body is `InstantiateWorkflowResponse`.
- `400` — the seed event failed schema validation.
- `404` — unknown `workflow_name`.
- `500` — catalog inconsistency: the seed builder found the catalog `Workflow.name` did not match the resolved `workflow_name` (a server-side data-integrity fault, not client input).
- `503` — store not configured / closed.
- `504` — the event was accepted and queued but the in-memory merge did not finish in time. **Each `POST` creates a new instance**, so do not blindly retry; poll the instances list or `GET` the instance state first.

but the OpenAPI spec only mentions `200`.

misi-bp:
# Description

Lungo's Agentic Workflows OpenAPI contract previously documented mostly success responses (for example `200` on instantiate), while FastAPI handlers can return a full set of 4xx/5xx codes (auth, validation, catalog errors, store unavailable, merge timeout, and others). Clients and tooling that rely on `schema/openapi/` therefore saw an incomplete API surface.

This change aligns the Lungo OpenAPI spec with reachable HTTP status codes from the backend, adds shared error response components, and locks the contract in CI so it does not drift again. **`schema/openapi/` is the source of truth** for paths, shapes, and status codes (edited manually or via LLM, not generated from Python).

**OpenAPI / schema**

* Add shared models and reusable responses in `components/schemas.yaml`: `HttpError`, `RequestValidationErrorItem`, and responses such as `Unauthorized`, `UnprocessableEntity`, `BadRequest`, `NotFound`, `InternalServerError`, `ServiceUnavailable`, and `GatewayTimeout`.
* Declare global API key Bearer security (`WorkflowApiKeyBearer`) in `openapi.yaml` so **401** applies to all operations without repeating it on every path.
* Extend `paths/agentic-workflows.yaml` so each operation declares the error (and non-200 success) statuses the backend can return, including **422** via `$ref` to `UnprocessableEntity` where path/query/body validation applies.

**Tests**

* Add `tests/unit/openapi/openapi_spec_helpers.py` to load the resolved spec and collect declared vs implemented status codes.
* Replace manifest-based drift checks with `test_agentic_workflow_openapi_status_codes.py`: verifies global Bearer `security`, every `operationId` declares `responses`, and handler literal `status_code=` values are a subset of OpenAPI-declared codes (plus **422** for FastAPI validation).

**Agent skills (ongoing maintenance)**

* Rename documentation skill to `agentic-workflows-api-documentation-lungo` (OpenAPI YAML + `workflow-instance_api.md` only; handler alignment via `openapi-to-python-lungo`).
* Update `openapi-to-python-lungo` and `reference.md` for OpenAPI-first status maintenance and cross-skill references.
* Refresh `AGENTS.md` skill index entries accordingly.

**Review follow-up**

* `HttpError` body-shape coupling (422 validation array vs string `detail` on other errors): deferred to a follow-up after #728 (frontend OpenAPI type codegen).

## Issue Link

Fixes #694

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [x] Other (please describe) — **OpenAPI contract enhancement** for Lungo Agentic Workflows (spec + CI drift gate + agent skill process); no runtime API behavior change.

## Checklist

- [ ] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [ ] I have verified this change is not present in other open pull requests
- [x] Functionality is documented (OpenAPI responses + agent skills; human API guide unchanged on this branch)

### Suggested verification for reviewers

From `coffeeAGNTCY/coffee_agents/lungo/`:

```bash
uv sync --extra dev
uv run pytest tests/unit/openapi/ -q
```


delthazor (dismissed):
lgtm

AGENTS.md
```diff
@@ -13,9 +13,9 @@
 | Topic | File |
 |-------|------|
 | Generate release notes | [.agents/skills/generate-release-notes/SKILL.md](.agents/skills/generate-release-notes/SKILL.md) |
-| OpenAPI → Python (lungo) | [.agents/skills/openapi-to-python-lungo/SKILL.md](.agents/skills/openapi-to-python-lungo/SKILL.md) |
+| OpenAPI → Python (lungo) | [.agents/skills/openapi-to-python-lungo/SKILL.md](.agents/skills/openapi-to-python-lungo/SKILL.md) — routers/DTOs; OpenAPI unit tests |
 | JSON Schema → Pydantic (lungo) | [.agents/skills/jsonschema-to-pydantic-lungo/SKILL.md](.agents/skills/jsonschema-to-pydantic-lungo/SKILL.md) |
-| API documentation (lungo) | [.agents/skills/api-documentation-lungo/SKILL.md](.agents/skills/api-documentation-lungo/SKILL.md) |
+| API documentation (lungo) | [.agents/skills/api-documentation-lungo/SKILL.md](.agents/skills/api-documentation-lungo/SKILL.md) — workflow-instance_api.md + OpenAPI error responses (#694) |
```
mihaialexandrescu: I don't think this is just about "error responses".

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -66,16 +69,17 @@ Copy this checklist and tick items as you go:
 - [ ] 1. Read every input in the Source registry (specs first, then router for behavior)
 - [ ] 2. Enumerate endpoints from OpenAPI and $defs from event_v1.json (do not hard-code)
 - [ ] 3. Reconcile spec against router.py; capture behavior the spec can't (the full reachable status-code set + trigger conditions from handlers, dependencies, and helper modules; SSE/NDJSON framing; backpressure)
-- [ ] 4. Write/refresh each document section per "Document structure", deriving every shape from a source
-- [ ] 5. Use real example values from the catalog data files; link (don't inline) full example payloads
-- [ ] 6. Apply the writing conventions
-- [ ] 7. Verify links and lint
-- [ ] 8. If new issues/PRs/files were referenced, update the Source registry AND the document (see "Keeping current")
+- [ ] 4. Update OpenAPI responses when the reachable status set changes (see "OpenAPI status codes (#694)")
```
mihaialexandrescu: I don't think we should mention the ticket of this particular PR on this line. This ticket will become irrelevant.

The other mentions of tickets were there because they outline an intent of what the sources of information should be and what is expected to be documented. This ticket flags a problem - it doesn't give long term instructions.

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -13,8 +13,10 @@ The lungo project root is `coffeeAGNTCY/coffee_agents/lungo/`. Paths in the [Sou
 ## Output
 
 - Document: `coffeeAGNTCY/coffee_agents/lungo/docs/workflow-instance_api.md`.
+- OpenAPI (when status/behavior changes): `schema/openapi/paths/agentic-workflows.yaml` — per-operation `responses` and descriptions; reuse `components/responses` in `schema/openapi/components/schemas.yaml`.
+- Curated drift manifest (when the reachable status set changes): `tests/unit/openapi/fixtures/agentic_workflows_expected_status_codes.yaml`.
 
-This document satisfies the interface-definition tickets it was first written for: [#447 — API endpoints](https://github.com/agntcy/coffeeAgntcy/issues/447) and [#446 — workflow-instance state JSON schema](https://github.com/agntcy/coffeeAgntcy/issues/446). Keep those links in the document's intro and schema section.
+This document satisfies the interface-definition tickets it was first written for: [#447 — API endpoints](https://github.com/agntcy/coffeeAgntcy/issues/447) and [#446 — workflow-instance state JSON schema](https://github.com/agntcy/coffeeAgntcy/issues/446). Keep those links in the document's intro and schema section. Status-code alignment with OpenAPI is tracked in [#694](https://github.com/agntcy/coffeeAgntcy/issues/694).
```
mihaialexandrescu: I don't think we should mention the ticket of this particular PR on this line. This ticket will become irrelevant.

The other mentions of tickets were there because they outline an intent of what the sources of information should be and what is expected to be documented. This ticket flags a problem - it doesn't give long term instructions. 

With the fixes provided in this PR, those other long term goals set in the other tickets should still hold.

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -66,16 +69,17 @@ Copy this checklist and tick items as you go:
...
+- [ ] 6. Use real example values from the catalog data files; link (don't inline) full example payloads
+- [ ] 7. Apply the writing conventions
+- [ ] 8. Verify links, lint, and OpenAPI unit tests (see Verification)
+- [ ] 9. If new issues/PRs/files were referenced, update the Source registry AND the document (see "Keeping current")
 ```
 
 ### Source-of-truth rules
 
-- **Derive, never invent.** Every endpoint, field, status code, and constraint must trace to a source file. If the spec and the router disagree, document the router's actual behavior and flag the drift to the user.
+- **Derive, never invent.** Every endpoint, field, status code, and constraint must trace to a source file. If the spec and the router disagree on **behavior**, treat **`router.py`** (and its dependencies/helpers) as authoritative for what the server returns; **update OpenAPI and the status manifest** so the contract catches up — do not leave OpenAPI under-documenting errors while only fixing prose.
```
mihaialexandrescu: I don't know that we want to treat `router.py` as **authoritative** in ALL cases. 
The next bullet point in the Skill definition treats the case of "things the spec omits".
The bullet point after that one explains how the code should complement and complete the view given by the OpenAPI spec (not making the code authoritative).
As a matter of principle, I think the point of these Skills is in fact to NOT rely on the code, but on the OpenAPI spec as a source of truth.

A simpler idea is to not consider a single source of truth and use the Skill to watch for drift. Then it would need to be a human operator's choice in a case by case basis which is correct (code or spec). I don't think we can always necessarily say that whichever has more status code is correct.

@pregnor, what do you think ?

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -102,13 +120,22 @@ done
...
 2. **Update the document** — re-run the [Workflow](#workflow) so every affected section reflects the new contract. Add or update issue/PR links where they belong (intro for epics, schema section for schema tickets, the relevant endpoint/section otherwise).
-3. **Update this skill** — if a new authoritative file/folder is now part of the contract surface, add a row to the [Source registry](#source-registry); if a new endpoint family or document section is needed, extend [Document structure](#document-structure). Keep the frontmatter `description` trigger terms in sync with any newly referenced paths so discovery stays accurate. Do not record one-off ticket numbers in the registry — link them from the document instead, and only add durable file/folder sources here.
-4. **Keep the index in sync** — this skill is registered in the repository [`AGENTS.md`](../../../AGENTS.md) Skills table; update that entry if the skill's name or scope changes.
+3. **Update OpenAPI and manifest** — if behavior or status codes changed, follow [OpenAPI status codes (#694)](#openapi-status-codes-694); do not document new errors only in markdown.
```
mihaialexandrescu: I don't think the Skill should follow the ticket behind this PR. This ticket flags an issue, does not provide long term instructions or sources of truth.

misi-bp: Sure it shouldn't, I remove all of them.

.agents/skills/api-documentation-lungo/SKILL.md
pregnor: I agree, the spec should be the source of truth.

misi-bp: As you have seen it already in https://github.com/agntcy/coffeeAgntcy/pull/728/changes, I'm also planning to use openapi spec as ssot, I just didn't want to deal with this in this ticket. Let me fix it now so this branch will also support openapi spec as ssot.

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -1,37 +1,38 @@
...
+  Authors and maintains the human-facing Agentic Workflows API documentation for the lungo subproject at coffeeAGNTCY/coffee_agents/lungo/docs/workflow-instance_api.md, and keeps hand-maintained OpenAPI in coffeeAGNTCY/coffee_agents/lungo/schema/openapi as the HTTP contract (paths, shapes, response statuses). Covers the catalog API, workflow-instance lifecycle/state, internal event ingress, SSE and NDJSON streaming, event_v1, and topology/use-case shapes. Use when documenting or updating the lungo workflow API; when OpenAPI under coffeeAGNTCY/coffee_agents/lungo/schema/openapi, JSON Schema under coffeeAGNTCY/coffee_agents/lungo/schema/jsonschemas, or api/agentic_workflows changes; or when a pull request changes lungo API contracts. DO NOT TRIGGER AUTOMATICALLY. ASK THE USER IF THE SKILL SHOULD BE USED.
```
mihaialexandrescu: > ; or when a pull request changes lungo API contracts.

I don't know that referencing only pull requests here in this portion of the Skill is the way to go. We sometimes do reference PRs in the prompts to the coding agents but the previous phrasing I think covered more of the common uses.

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -63,31 +64,49 @@ Keep the document organized in this order (the headings are stable so links and
...
-- **Enumerate, don't count.** Build the endpoint summary and `$defs` table by listing what's in the specs so the document stays correct as endpoints/types are added.
-- **Real examples.** Pull patterns, use-cases, workflow summaries, and topology from `patterns.py`, `use_cases.py`, and `starting_workflows.json` so samples match what the API returns.
+- **OpenAPI first.** Paths, request/response shapes, and declared HTTP status codes live in `schema/openapi/`. Change the spec first, then handlers, human doc, and generated types (frontend `npm run generate:api-types`, Python via openapi-to-python-lungo).
+- **Derive, never invent.** Every endpoint, field, and status code in the human doc must trace to OpenAPI (plus global `401` from `security`) or `event_v1.json`.
+- **Router implements the contract.** If handlers return a status OpenAPI does not declare, **add it to OpenAPI** or **change the handler** - do not document undeclared codes only in markdown.
```
mihaialexandrescu: > **add it to OpenAPI** or **change the handler**

While the first one is on the very edge of what I would consider reasonable for a Skill called `api-documentation`, the second part about making changes to the `handler` is not.

We may need to rename the Skill if we want it to change code.

Later edit:
We have the `openapi-to-python-lungo` Skill for code changes. It could be that "change the handler" might just call on that Skill but I am not convinced. Either way, code changes should be made by the other Skill and/or definitely not by a Skill called "documentation".

.agents/skills/openapi-to-python-lungo/SKILL.md
```diff
@@ -1,7 +1,7 @@
...
 description: >-
-  Generates and validates Python FastAPI routers and DTOs from the OpenAPI schemas in coffeeAGNTCY/coffee_agents/lungo/schema/openapi for the lungo subproject. Use when adding, removing, renaming, or modifying endpoints in any lungo OpenAPI document; when generating or regenerating files under coffeeAGNTCY/coffee_agents/lungo/api/<tag>/ (router.py, dtos.py); when validating that hand-edited handlers or DTOs still match the OpenAPI spec; or when scaffolding the OpenAPI unit tests for a router.
+  Generates and validates Python FastAPI routers and DTOs from the OpenAPI schemas in coffeeAGNTCY/coffee_agents/lungo/schema/openapi for the lungo subproject. When contracts change, the sibling api-documentation-lungo skill updates hand-maintained OpenAPI operation `responses` and `docs/workflow-instance_api.md`. Use when adding, removing, renaming, or modifying endpoints in any lungo OpenAPI document; when generating or regenerating files under coffeeAGNTCY/coffee_agents/lungo/api/<tag>/ (router.py, dtos.py); when validating that hand-edited handlers or DTOs still match the OpenAPI spec; or when scaffolding the OpenAPI unit tests for a router.
```
mihaialexandrescu: IF we end up renaming the `api-documentation-lungo` Skill, we should not forget to update it here in this file too (multiple occurences, commenting just once for the review).
Otherwise, feel free to resolve this thread/conversation at your convenience.

misi-bp: I have no experience to decide on this on my own - renaming for now seems to be a valid request, but I have had no idea so far what are the limits of each skill in this project.

mihaialexandrescu: You are right. It's meant as an invitation for other reviewers to chime in as well.

pregnor: IMO if the new boundary of the skill is only updating the docs because the generation handles the rest, then I would suggest the name `lungo-agentic-workflows-api-documentation` - if I recall right and this is only done for the AWAPI OpenAPI spec.

Otherwise I support the renaming with the narrowed scope and good catch by Mihai that in case of renaming we should be thorough.

Edit: And I agree if we change the "content" of the skill we should name it accordingly and we shouldn't name it documentation if there is code change attached, it's probably worth splitting into 2 skills in that case.

.agents/skills/api-documentation-lungo/SKILL.md
pregnor: **Recommended**: (should) Not sure if hand-maintained is the right phrase here, it's "the source of truth that might be updated manually or via LLM, but never code generated", that's the point.

Not sure what is the best phrasing, maybe @mihaialexandrescu can correct my hunglish, but I would aim for something along those lines.

.agents/skills/api-documentation-lungo/SKILL.md
```diff
@@ -1,37 +1,38 @@
...
 
 ## Output
 
 - Document: `coffeeAGNTCY/coffee_agents/lungo/docs/workflow-instance_api.md`.
+- OpenAPI (when paths, shapes, or response statuses change): `schema/openapi/paths/agentic-workflows.yaml` - per-operation `responses` and descriptions; reuse `components/responses` in `schema/openapi/components/schemas.yaml`.
 
-This document satisfies the interface-definition tickets it was first written for: [#447 - API endpoints](https://github.com/agntcy/coffeeAgntcy/issues/447) and [#446 - workflow-instance state JSON schema](https://github.com/agntcy/coffeeAgntcy/issues/446). Keep those links in the document's intro and schema section.
+The human guide and OpenAPI path items must stay aligned. Derive everything from the [Source registry](#source-registry). Reconcile the human doc using [Document structure](#document-structure) whenever the API surface changes. When HTTP statuses change, follow [OpenAPI status codes](#openapi-status-codes) **before** updating handlers.
```
pregnor: **Recommended**: IMO we **SHOULD** point the source of truth not to the source registry, but to the `schema/openapi/...` OpenAPI spec, because this markdown is descriptive, but not exhaustive from programming interface-wise, while the OpenAPI spec is.

We should update the source registry description here based on that and keep this as human friendly readable summary, but the spec will always be more detailed.

Which I do acknowledge may be iterated by LLM generated code not just manually by humans, hence my first comment above.

@mihaialexandrescu am I making sense here?

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/components/schemas.yaml
```diff
@@ -180,3 +180,102 @@ components:
...
+        HTTPException uses a string detail; validation failures use an array of items.
+      properties:
+        detail:
+          oneOf:
+            - type: string
+              description: Typical application error message (4xx/5xx from HTTPException).
+            - type: array
+              items:
+                $ref: "#/components/schemas/RequestValidationErrorItem"
+              description: Request validation errors (422).
```
pregnor, line 216: **Question**: Wouldn't be this cleaner if we didn't just define the 422 but also some placeholder for the rest of the error codes, or do we intend those to be dynamically constructed all the time?

My expectation would be to fix the HTTP status code returned at least for a NotFound schema type and allow custom string etc.

Alternatively I would be okay with omitting the `BadRequest`, `NotFound`, etc.  types, because then it's just freely dynamic, but I don't like defining those while not restricting the HTTP status code to their respective ones. (being able to create a NotFound type with 413 instead of 404).

I also see how you use it coupled at coffeeAGNTCY/coffee_agents/lungo/schema/openapi/paths/agentic-workflows.yaml, but if we make an error in the coupling there, it will be a somewhat hidden bug. I would prefer to have the object used there defined here (e.g. `{"422": "$ref: ../components/schemas.yaml#/components/responses/UnprocessableEntity"}`, not entirely sure if this is the syntax or it is possible simply with a yaml object like this, and then using a oneOf there or something, this needs investigation).

Alternatively if my suggestion doesn't work: https://github.com/cisco-eti/policy-framework/blob/main/policyport/api/restapi/policyport.gen.yaml#L937 (it's not a necessity to have 1 errorResponseSchema at the root, the point is the coupling in components already).

pregnor (commented):
Mihai's comments are a must for me.

I left a couple open ended questions that may necessitate some changes, but I'm not saying it's not mergeable today, I just have differing preferences until discussed.

.agents/skills/api-documentation-lungo/SKILL.md
mihaialexandrescu: > the source of truth that might be updated manually or via LLM, but never generated from code

is what I'd go for (just switched up the words in the last part in your suggestion).

.agents/skills/api-documentation-lungo/SKILL.md
mihaialexandrescu: I think in the `Source registry` section we can only reference the `schema/openapi/` and `schema/jsconschemas` folders and do away with mentioning in the Skill the **current** specific file paths inside those folders and with the descriptions of what was once in them (see the current "what to extract" column) -- those files or how they are broken up with imports may change substantially overnight and the Skill should remain generic throughout. LLMs are good enough right now to adapt to reading the files from scratch.

It's not how that section was written prior to this PR (so that "blame" is on me) but it may be better to make this update from the previous paragraph too.

I also think, we can do away, from that "Source Registry" section, with the precise references to Python code files (as 3 examples: `router.py` and `dtos.py`, `common/workflow_instance_store/interfaces.py`).
We have the other 2 Skills that deal with code generation from the specs.

I also don't think we should reference as a "Source" (of truth) files like `docs/a2a_event_schema_middleware.md`. 

We are talking in this thread mainly about changes to another section but the LLMs should deal with it ok.

.agents/skills/openapi-to-python-lungo/SKILL.md
mihaialexandrescu: For code, I reckon we can rely on the other 2 existing code generation skills: `jsonschema-to-pydantic-lungo` and `openapi-to-python-lungo`.

I that case we can keep `api-documentation-lungo` oriented towards documentation and then the only renaming is of that Skill to what Patrik suggested. In order to keep the current naming convention of the Skills I would put `lungo` at the end, though: `agentic-workflows-api-documentation-lungo`. 
About the naming convention reference:  Whether we like it or not it's what we have now and I think polluting this PRs (already balooning) scope with renaming the other skills would help anyone.

pregnor: I support Mihai's points.

misi-bp:
@pregnor @mihaialexandrescu 
Thank you for your comments above. Since those are more or less tied together, as a final result I reply to them here; this is my summary of changes done:

1. Rename api-documentation-lungo to agentic-workflows-api-documentation-lungo.
2. Limit that skill to OpenAPI YAML and workflow-instance_api.md; 
3. Route handler alignment to openapi-to-python-lungo.
4. Use folder-level SSOT registry
5. Restore broad description triggers, and replace hand-maintained wording with OpenAPI
as source of truth (edited manually or via LLM, not generated from code).

As for coupling HttpErrors, I left a comment in #728 to complete it in that PR once the generation from openapi spec is implemented and finalized.

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/components/schemas.yaml
pregnor: From Misi:
> As for coupling HttpErrors, I left a comment in https://github.com/agntcy/coffeeAgntcy/pull/728 to complete it in that PR once the generation from openapi spec is implemented and finalized.

Resolving this thread for now.

pregnor (dismissed):
From my perspective it's LGTM.

Mihai please check the outcomes for your comments.
The LLM said the read only python paths might be questionable, the rest should be done.

.agents/skills/api-documentation-lungo/SKILL.md
misi-bp: Following your comment, I went over the requested changes with AI, let me show the important parts of its answer. I have also taken into consideration what changes the #728 is to bring.

**How the three skills divide work (after simplification)**

agentic-workflows-api-documentation-lungo
  → schema/openapi/ + schema/jsonschemas/ + workflow-instance_api.md

openapi-to-python-lungo
  → openapi.yaml, paths/*, api/<tag>/router.py, dtos.py, tests

jsonschema-to-pydantic-lungo
  → schema/jsonschemas/* → schema/types/*

#728 (685/hygiene)
  → npm run generate:api-types after OpenAPI/JSON Schema changes

@pregnor , @mihaialexandrescu 
_Are we satisfied with these above? 
Are we OK to go with the caveats below?_

**Minor caveats (operational, not blockers)**

1. Agents need some edit hint
The reviewer targets the registry table, not every section. Keeping short operational notes in “OpenAPI status codes” (e.g. “edit path items under schema/openapi/; resolve from openapi.yaml”) is fine and helps agents find the right file without a 15-row inventory. (That does not conflict with #728.

2. Example/catalog data
patterns.py, starting_workflows.json are for examples in prose, not SSOT. You can genericize to “catalog data under api/agentic_workflows/” or drop from the registry; #728 doesn’t care.

3. Trigger-condition prose
If you remove the whole “Implementation context” Python list, agents may write vaguer “when X returns 404” bullets unless they still peek at api/agentic_workflows/ on demand. That’s a doc quality tradeoff, not a codegen break. openapi-to-python-lungo remains the place that knows those files.

4. Optional one-liner for discovery
Without listing every file, a single line helps: “Resolve the contract from schema/openapi/openapi.yaml and its $refs.” That aligns with #728’s script and prance, without a brittle registry.

coffeeAGNTCY/coffee_agents/lungo/schema/openapi/components/schemas.yaml
misi-bp: Short answer from AI:

Standard OpenAPI cannot do status-keyed maps in components. Coupling has to be:

- at the operation ("404": …), and/or
- via different body schemas per response component, and/or
- via custom lint (Spectral, pytest) that checks naming conventions.

The policyport pattern (could not fetch the file; repo may be private) is usually: per-status or per-error-kind body schemas with names that codegen understands, still referenced from operation keys. The “coupling in components” is body shape, not HTTP code inside the component.

The YAML idea ({"422": "$ref": ...} in components is not valid OpenAPI 3.

Some parts/ideas of your comments should and can be implemented, but recommended to do so only after #728 is merged. See my 2nd follow-up comment there: https://github.com/agntcy/coffeeAgntcy/pull/728#issuecomment-5177210759

.agents/skills/openapi-to-python-lungo/SKILL.md
```diff
@@ -61,7 +63,8 @@ Track progress with this checklist:
 - [ ] 3. Regenerate dtos.py from scratch
 - [ ] 4. Reconcile router.py (preserve handler bodies)
 - [ ] 5. Generate or refresh tests under tests/unit/openapi/
-- [ ] 6. Run the tests and the linter
+- [ ] 6. If contracts changed, ensure OpenAPI and `workflow-instance_api.md` are updated ([agentic-workflows-api-documentation-lungo](../agentic-workflows-api-documentation-lungo/SKILL.md)); align handlers to use only OpenAPI-declared statuses
```
mihaialexandrescu, line 66: I have some doubts about how this will play out in the following workflow (I don't mean a CoffeeAgntcy workflow) scenario:  we want to prototype some code (with the intention to eventually add to the OpenAPI spec the changes that stick) and the LLM adds the Python code but with some new status codes and we don't want this Skill to come and bulldoze over that prototype.

I don't know for sure if this will become a problem so I'm not going to hold this up over it. If we see it become a problem, we'll address it then.

mihaialexandrescu (approved):

.agents/skills/api-documentation-lungo/SKILL.md
pregnor: looks ok to me if the first component of the line is always the input, seems like so.

pregnor (approved):

PR #727, opened 2026-07-29
fix(corto,lungo): fix transport env var lazy load

pregnor:
# Description

The transform env loading expected the env vars to be preloaded because it selectively imported the loaded env vars instead of the full config module that would do the dotenv loading.
This affected local runs.
I changed it so we are loading the full module so dotenv is loaded.
We have to revise the config interface to make this less error-prone in the future.

## Issue Link

Link the primary issue in the PR description using `#` (e.g. `Fixes #123`). This enables two‑way linking.

## Type of Change

- [x] Bugfix

## Checklist


mihaialexandrescu (approved):

delthazor (approved):
lgtm

PR #733, opened 2026-08-05
fix: trigger python tests on PRs and push events

### LINKED ISSUE #734: Fix Python tests not triggering in CI after split
After (and during) https://github.com/agntcy/coffeeAgntcy/pull/707, the Python tests no longer run in CI due to a workflow creation error.
Example workflows from right before the merge of that PR: https://github.com/agntcy/coffeeAgntcy/actions/runs/30627478621

The same keeps happening after that change made it into `main`.

The Python Tests workflow fails at load time on `push` to `main` and on `pull_request`, so no subproject Python tests run.

`test.yaml` passes `${{ inputs.test_corto }}` / `test_lungo` / `test_recruiter` straight into the `test-subprojects-reusable.yaml` reusable workflow. The `inputs` context only exists for `workflow_call` / `workflow_dispatch`, so on `push` and `pull_request` those expressions rendered to an empty string `''`. Since the reusable workflow declares those inputs as `type: boolean`, `''` is not a valid value and GitHub rejects the run during "load reusable workflow context":
```
.github/workflows/test.yaml (Line: 103, Col: 19): Unexpected value ''
```
Because every tests / * job needs the `subprojects-to-test` job, the whole run fails before any job can be scheduled — the tests appear to never trigger.

### LINKED ISSUE #735: Python tests fail in CI for all 3 subprojects
After PR  #707 some Python tests in all 3 subprojects fail in CI.
Example: https://github.com/agntcy/coffeeAgntcy/actions/runs/31006888754/job/92308938841

Usually they fail to set up `slim`.
Example from Corto:
```
_ ERROR at setup of test_get_agent_oasf_returns_200_and_valid_json[exchange-supervisor-agent] _

    @pytest.fixture(scope="session", autouse=True)
    def orchestrate_session_services():
        print("\n--- Setting up session level service integrations ---")
>       setup_transports()

tests/integration/conftest.py:83: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/integration/conftest.py:92: in setup_transports
    _startup_slim()
tests/integration/conftest.py:104: in _startup_slim
    up(files, ["slim"])
tests/integration/helpers/docker_helpers.py:71: in up
    wait_for_service(files, svc)
tests/integration/helpers/docker_helpers.py:133: in wait_for_service
    cid = _container_id(files, service)  # ensures same compose context
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

files = ['docker-compose.yaml'], service = 'slim'

    def _container_id(files: List[str], service: str) -> str:
        cmd = _compose_cmd(files) + ["ps", "-a", "-q", service]
        res = subprocess.run(cmd, capture_output=True, text=True, cwd=PROJECT_DIR)
        cid = res.stdout.strip()
    
        if res.returncode == 0 and cid:
            return cid
    
        print(f"\nFailed to find container ID for service '{service}\n")
        ps_out = subprocess.run(
            _compose_cmd(files) + ["ps", "--format", "table"],
            capture_output=True, text=True, cwd=PROJECT_DIR
        )
        print("docker compose ps output:")
        print(ps_out.stdout or ps_out.stderr or "<no output>")
>       raise RuntimeError(
            f"No container id found for service '{service}'."
        )
E       RuntimeError: No container id found for service 'slim'.

... ... ...
```

mihaialexandrescu:
# Description

The Python Tests workflow was failing at load time on push to main (and on `pull_request`), so no subproject tests ran. This PR fixes the invalid reusable-workflow inputs and removes a dead branch condition.

`test.yaml` passed `${{ inputs.test_corto }}` / `test_lungo` / `test_recruiter` straight into the `test-subprojects-reusable.yaml` reusable workflow. The `inputs` context only exists for `workflow_call` / `workflow_dispatch`, so on `push` and `pull_request` those expressions rendered to an empty string `''`. Since the reusable workflow declares those inputs as `type: boolean`, '' is not a valid value and GitHub rejected the run during "load reusable workflow context":
```
.github/workflows/test.yaml (Line: 103, Col: 19): Unexpected value ''
```
Because every tests / * job needs the `subprojects-to-test` job, the whole run failed before any job was scheduled — the tests appeared to never trigger.

### Changes
  * `test.yaml`: coerce the passthrough inputs to a valid boolean with `${{ inputs.<x> || false }}`, and add a comment explaining why. The reusable workflow ignores these values on `push`/`pull_request` (it branches on `github.event_name`), so `|| false` only ever applies to dispatch/call runs, where it still correctly honors an explicit `false`.
    * I chose the shorter version plus an educational comment but we could also have a longer and more explicit form of that condition (example for Corto below):
      ```
      test_corto: ${{ (github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call') && inputs.test_corto || false }}
      ```
  * `test-subprojects-reusable.yaml`: drop the dead `github.event_name == 'workflow_call'` condition. Inside a reusable workflow `github.event_name` is the **originating event** (`push` / `pull_request` / `workflow_dispatch`) and is never `workflow_call`, so that condition could never match.
    * From https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_call: 
      > `workflow_call` is used to indicate that a workflow can be called by another workflow. When a workflow is triggered with the `workflow_call` event, the event payload in the called workflow is the same event payload from the calling workflow.

  * the rest are fixes for the Python tests that we didn't see before due to the previous issues
  
## Issue Link

Fixes #734
Fixes #735 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/workflows/test-subprojects-reusable.yaml
+++ b/.github/workflows/test-subprojects-reusable.yaml
@@ -116,7 +116,7 @@ jobs:
             test_corto=true
             test_lungo=true
             test_recruiter=true
-          elif [[ "${{ github.event_name }}" == "workflow_call" || "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+          elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
             test_corto=${{ inputs.test_corto }}
             test_lungo=${{ inputs.test_lungo }}
             test_recruiter=${{ inputs.test_recruiter }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -100,9 +100,14 @@ jobs:
   subprojects-to-test:
     uses: ./.github/workflows/test-subprojects-reusable.yaml
     with:
-      test_corto: ${{ inputs.test_corto }}
-      test_lungo: ${{ inputs.test_lungo }}
-      test_recruiter: ${{ inputs.test_recruiter }}
+      # `inputs` only exists for workflow_call / workflow_dispatch; on push and
+      # pull_request it is empty, so these render to '' (invalid for the boolean
+      # inputs). `|| false` coerces them to a valid boolean. The reusable workflow
+      # ignores these values on push/pull_request anyway (it branches on
+      # github.event_name), so `false` here only ever applies to dispatch/call.
+      test_corto: ${{ inputs.test_corto || false }}
+      test_lungo: ${{ inputs.test_lungo || false }}
+      test_recruiter: ${{ inputs.test_recruiter || false }}
 
   tests-corto:
     name: tests / corto
--- a/coffeeAGNTCY/coffee_agents/corto/docker-compose.yaml
+++ b/coffeeAGNTCY/coffee_agents/corto/docker-compose.yaml
@@ -11,7 +11,8 @@ services:
     image: ghcr.io/agntcy/coffee-agntcy/corto-farm:latest
     container_name: farm-server
     env_file:
-      - .env
+      - path: .env
+        required: false
     environment:
       - FARM_AGENT_HOST=farm-server
       - FARM_AGENT_PORT=9999
@@ -59,7 +60,8 @@ services:
     image: ghcr.io/agntcy/coffee-agntcy/corto-exchange:latest
     container_name: exchange-server-corto
     env_file:
-      - .env
+      - path: .env
+        required: false
     environment:
       - FARM_AGENT_HOST=farm-server
       - FARM_AGENT_PORT=9999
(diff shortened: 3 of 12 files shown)


pregnor (approved):

misi-bp (approved):
The code lgtm, also tested succesfully WITHOUT VPN - for some reason, via VPN the containers do not start. If this is expected and the reason is out of scope, the the PR is approved from my end.

mihaialexandrescu:
For me, there was one day last week when the build failed while on the VPN but was ok without the VPN.
The last 2 days, both were fine in my case.

PR #748, opened 2026-08-24
Fix test flakiness with Weather MCP improvements

### LINKED ISSUE #373: [Bug]: Integration test failing with unexpected response from weather MCP server
### Bug Description

One of the integration tests is failing when run as a gh workflow. The weather MCP server is not returning valid weather which causes an error response from the Colombia agent when a integer amount is expected.

### Expected Behavior

The Colombia agent should handle scenarios where weather can not be retrieved and still return a quantity. 

### Affected Version

v1.0.0

### Steps to Reproduce

I have not been able to replicate via local integration tests but gh workflow integration tests reproduce the issue.

### Checklist

delthazor:
# Description
CI test were sometimes flaky due to Weather MCP returning a non-valid weather response (most commonly an error message)

To solve this, the following steps were made:
 - Remove named location coord lookup, as it used another 3rd party service that could potentially timeout, or return an error, even if everything was ok on our side
 - Introduce a stub for open-meteo 3rd party live service, so tests no longer rely on an external API that can timeout, or be unreachable, or return unexpected responses. Test therefore now use weather MCP with a stubbed internal API returning controlled weather data.
 - Weather MCP no longer returns success, when there is an error message in the response. In all cases it raises an error, and only returns a successful response if there is a proper whether data to be sent back.
- New config flag was introduced to determine how we want to handle Weather MCP errors. Since during development we might want errors to cause real failures, so they don't go unnoticed, the default value for the flag is to report the failure. In some environments we might want to gracefully recover, by silently discarding the error, and use a pre-determined weather scenario. For this the flag can be set to use said fallback.

Also, corresponding tests were updated, so they cover the new functionality, testing both states of the flag, and relying on the new stub.

## Issue Link
Fixes #373 

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist


mihaialexandrescu (dismissed):

pregnor (dismissed):

delthazor:
Rebased on main with conflict resolution

mihaialexandrescu (approved):

pregnor (approved):

PR #753, opened 2026-09-07
ci: + ci-gate summary workflow

pregnor:
# Description

Introduced a new CI workflow called `CI Gate` that has 2 major goals:
- Validate the workflows by running some GHA workflow linting on them using `reviewdog/action-actionlint` to prevent issues arising from the workflow not starting due to syntactic or structural issues.
- Collect all workflow results and give a boolean result for the CI package.

The latter is intended to be used as a flexible required status check that we could wire into the branch protection to both improve our quality gate, mark more scorecard fields and meanwhile keep control over what constitutes the required status check in an in-repo way.

Right now all running CI jobs are expected to be green (even newly added ones or removals don't require any change), later on we can transition to allowlist/disallowlist if needed.

Also fixed some linter issues.

## Issue Link

.

## Type of Change

- [X] Other (please describe): CI

## Checklist

### DIFF
--- a/.github/workflows/ci-gate.yaml
+++ b/.github/workflows/ci-gate.yaml
@@ -1,0 +1,175 @@
+name: CI Gate
+
+# Single required status check for the repo.
+#
+# - Validates every workflow file under .github/workflows/ (schema, syntax,
+#   structure) via actionlint, on every run - no manual list to keep in sync.
+# - Waits for every sibling workflow run triggered on the same commit to reach
+#   a terminal state, by polling the Actions runs API (not the checks API -
+#   that one silently omits startup_failure runs), then reports on all of them,
+#   including ones that failed at startup, were skipped, or never completed.
+# - Never short-circuits: validation and the wait/report step both run
+#   regardless of each other's outcome, and the pass/fail decision is made
+#   once everything has been collected.
+
+on:
+  pull_request: {}
+  push:
+    branches: [main]
+    tags: ["*"]
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+  actions: read
+
+env:
+  SETTLE_DELAY_SECONDS: 60
+  POLL_INTERVAL_SECONDS: 60
+  OVERALL_TIMEOUT_SECONDS: 6000 # 100 minutes
+  CONSECUTIVE_CLEAN_POLLS_REQUIRED: 2
+
+jobs:
+  ci-gate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 105
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
+
+      - name: Validate all workflow files (schema, syntax, structure)
+        id: validate
+        continue-on-error: true
+        uses: reviewdog/action-actionlint@d290e336d5a743810aef4404f757dc862276d2ae # v1.73.4
+        with:
+          reporter: local
+          fail_level: error
+          filter_mode: nofilter
+
+      - name: Wait for sibling workflow runs on this commit
+        id: wait
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -uo pipefail
+
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            head_sha="${{ github.event.pull_request.head.sha }}"
+          else
+            head_sha="${{ github.sha }}"
+          fi
+
+          self_run_id="${{ github.run_id }}"
+          start_time=$(date +%s)
+          clean_polls=0
+          settled=false
+          last_good_runs_json='[]'
+
+          echo "Waiting ${SETTLE_DELAY_SECONDS}s for sibling runs to be scheduled..."
+          sleep "$SETTLE_DELAY_SECONDS"
+
+          while true; do
+            if fetched=$(gh api "repos/${{ github.repository }}/actions/runs?head_sha=${head_sha}&per_page=100" \
+              --paginate --jq '.workflow_runs[]' | jq -s '.'); then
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-gate.yaml
```diff
@@ -0,0 +1,175 @@
...
+          echo "Waiting ${SETTLE_DELAY_SECONDS}s for sibling runs to be scheduled..."
+          sleep "$SETTLE_DELAY_SECONDS"
+
+          while true; do
+            if fetched=$(gh api "repos/${{ github.repository }}/actions/runs?head_sha=${head_sha}&per_page=100" \
+              --paginate --jq '.workflow_runs[]' | jq -s '.'); then
+              last_good_runs_json="$fetched"
+
+              pending=$(jq --argjson self "$self_run_id" \
+                '[.[] | select(.id != $self) | select(.status != "completed")] | length' <<<"$last_good_runs_json")
```
delthazor, line 80: Won't this deadlock if parallel runs are started in quick succession?
I'm not really familiar with this, but it seems like it would count not just sibling workflows, but multiple instances of the same workflow too.

.github/workflows/ci-gate.yaml
```diff
@@ -0,0 +1,175 @@
...
+          start_time=$(date +%s)
+          clean_polls=0
+          settled=false
+          last_good_runs_json='[]'
+
+          echo "Waiting ${SETTLE_DELAY_SECONDS}s for sibling runs to be scheduled..."
+          sleep "$SETTLE_DELAY_SECONDS"
+
+          while true; do
+            if fetched=$(gh api "repos/${{ github.repository }}/actions/runs?head_sha=${head_sha}&per_page=100" \
```
delthazor, line 75: Does this query all runs for the given sha? Even the superseded ones? How are cancelled runs supposed to be treated?

delthazor (commented):
You may also consider updating the workflows readme file

.github/workflows/ci-gate.yaml
pregnor: 1. It checks the CI job runs (instances of CI workflows) that are actively running on the commit SHA, that's the `${head_sha}` part in the query, yes.
2. Runs cannot be superseded for a SHA, because commit hashes are unique and commits are immutable in the sense that the same commit hash cannot disguise different content, that's an invariant of git. If you push a new commit to a branch - or amend/rebase a commit - the `head_sha` will change and the query will only list the latest CI job runs (instances), not previous ones.
In practice this means that the check only checks the last commit state (but that compounds earlier commit deltas in a single resulting code state).
Edit: this is the same behavior we have today, the checks run per commit, plus on PR level.
3. There are 2 run metadata controlling the end result, `status` and `conclusion`.
Here https://docs.github.com/en/pull-requests/reference/status-checks#check-statuses-and-conclusions you can see the available `status` values which is the answer to "where is the workflow in it's lifecycle". 
Right after on the same page it lists the `conclusion` values edit: that answers the "how did the run finish?" question.
For a cancelled run the metadata will be `status: completed, conclusion: cancelled`.
This particular step only waits for pending jobs to reach any kind of final state (`completed`, `failure`, `startup_failure` AFAIK).
Afterwards the follow-up step will examine the `conclusion` and `success` and `skip` is expected of all workflows for this summary to be succeeding, every other kind of conclusion fails it.

pregnor: Otherwise the waiting step will just time out on this workflow and that fails the gate.

Also my understanding from compounding sources is that whenever the workflow YAML we write does not succeed, the `status` will always be `completed` with various `conclusion`s based on the content.
The rest of the final states `failure, startup_failure` are for the framework failures when the GHA execution fails, not the workflow content.
For example if you make a syntactically invalid `workflow.yaml` it will be `startup_failure` status.

The point is we require 2 things for a passing `CI Gate`:
1. every active workflow for the current state should be in `status: completed` (line 80), AND
2. every active workflow for the current state should have `conclusion: success | skipped` (line 146)

If any of these conditions is not met then we fail the CI Gate, including inability to conclude the gate in 100m.

.github/workflows/ci-gate.yaml
pregnor: It would, if we were checking for runs in general.
With my explanation on per (immutable) commit SHAs, there will only ever be 1 CI Gate that is checking that particular state's runs and because of that there will never be 2 CI Gate runs seeing each other and thus waiting on each other.
Every CI trigger should only ever trigger 1 CI Gate run which will wait for all other active runs on the SHA to finish.

If we ever needed to be able to have multiple CI Gate runs watch the same runs then we just need to change the `run_id` based check to be a workflow ID based one which does not change between invocations, but I don't see the need for that just yet.

If we ever run into deadlocks there is always manual cancellation possible on top of the auto-timeout feature and we can resolve it afterwards by fixing the workflows.

pregnor: Also just to be sure double checked whether the SHA approach covers the PR triggered workflows as well, and according to the LLM it does:

> And querying the runs API directly for that same head commit confirms all 6 runs (CI Gate + 5 siblings: Scorecard, Source lint, Helm Push Charts, Python Tests, Build and Push Project Docker Images) are indexed under event: "pull_request" with the correct shared head_sha — exactly the mechanism the design relies on.
> 
> So there's nothing to enable — PR coverage was already part of the original design and is now confirmed working in production, not just in the dry-run test I did earlier against historical data. The push: trigger being scoped to branches: [main] + tags: ["*"] (rather than all branches) is intentional, matching every other workflow in this repo — nobody expects full CI to run on a bare feature-branch push before a PR exists.

pregnor:
> You may also consider updating the workflows readme file

I pushed the update

mihaialexandrescu (approved):

delthazor (approved):
I got sufficient explanation, approved.

PR #768, opened 2026-09-18
feat(corto,lungo): default SLIM timeout, reduce retry count, surface real backend errors in frontend

mihaialexandrescu:
# Description

This PR aims to bring up to date 2 older community PRs that had significant overlap and in the mean time also have merge conflicts:
  * https://github.com/agntcy/coffeeAgntcy/pull/579/
  * https://github.com/agntcy/coffeeAgntcy/pull/520/

Credit still goes to the initial contributors.

There are also some new additions regarding error handling in the frontend code and some refinements regarding the passing of timeouts to SLIM.

I ran the containers locally and tested coffeeAgntcy in the UI with a few suggested prompts.

I ran locally the test suite so that it includes the LLM-dependent tests as well for Lungo and all of them passed (555/555).

```bash
|> % uv run --no-sync pytest tests
...
collected 555 items

tests/integration/test_auction.py ...                                  [  0%]
tests/integration/test_logistics_accountant.py .                       [  0%]
... ...
tests/integration/llm/test_auction_flows.py ..................         [  5%]
tests/integration/llm/test_colombia_weather_fallback.py ....           [  6%]
tests/integration/llm/test_logistics_supervisor_flows.py ..            [  6%]
tests/integration/llm/test_pattern_chat_proxy.py .                     [  6%]
... ...
tests/unit/schemas/types/test_event.py ..................              [100%]

===================== 555 passed in 801.13s (0:13:21) =======================
```

## Issue Link

Fixes https://github.com/agntcy/coffeeAgntcy/issues/519 and https://github.com/agntcy/coffeeAgntcy/issues/518

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have read the [contributing guidelines](/agntcy/coffeeAgntcy/blob/main/CONTRIBUTING.md)
- [x] Existing issues have been referenced (where applicable)
- [ ] I have verified this change is not present in other open pull requests
- [ ] Functionality is documented
- [x] All code style checks pass
- [x] New code contribution is covered by automated tests
- [x] All new and existing tests pass

PR #776, opened 2026-09-21
747/python indentation

### LINKED ISSUE #747: [Feature]: Clarify python indentation standard
### Problem Statement

Currently, most of the project is using 4 paces as indentation for python files, but some files (especially in the middleware section) uses tabs instead.

This is an inconsistency in and of itself, that should be clarified for contributors.

But in addition this can cause mixed files, when code parts are moved around during refactors. Also, coding agents can be confused and sometimes generate mixed indentation files, or append code with different indentation to existing files. (one such example occurred in this PR: https://github.com/agntcy/coffeeAgntcy/pull/745)

It worth to mention that the JS/TS part of the code (frontend side) already has formatting rules for this exact (and many more) issue(s) thanks to the prettier/eslint package

### Proposed Solution

Must have:
-Clarify a single indentation format as project standard, that is stated in the documentation
-Transform all files that does not comply with the accepted convention
-Block mixed indentation files by a CI validation step (e.g. using -tabnanny python switch and check for errors)

Recommended:
-Check or even enforce the agreed indentation standard with a CI verification step (with tools like black or ruff)

Optional:
-Automate formatting (with editorconfig or linter tools like ruff)

### Alternatives Considered

Alternatives solution only applies to the verification (or enforcement) steps.

The following alternatives were identified:

 # Option 1: tabnanny (stdlib only)
What it is: A built-in Python module that scans source files for ambiguous or inconsistent indentation like mixing tabs and spaces in ways Python might misinterpret.

Pros:
 - Zero new dependencies
 - Zero license/vendor risk
 - Fast, trivial to add as a CI step

Cons:
 - Detection only — does not auto-fix or format
 - Does not enforce the selected style if a file is consistent
 - Weak as a long-term engineering standard compared to a formatter

# Option 2: Ruff (lint + format)
What it is: One fast tool (Astral, same ecosystem as uv) that replaces much of flake8 + isort + Black (used as previous industry standard ). Widely adopted in modern Python projects.

Pros:
 - Strong enforcement: format + lint in one tool
 - ruff check can flag tab-related issues and hundreds of other quality rules
 - Fits existing stack: uv already in CI (setup-uv, uv sync)
 - Single dev dependency; low operational overhead
 - Can take care of the migration step (auto format files with different indentation)
 - Can be rolled out gradually (e.g. tests/ first, then agents/, then all)

Cons
 - Requires adding ruff to dev dependencies
 - Might need a phased rollout plan
 - Can be more heavy-weight than strictly needed

# Option 3: Black with optional extra tools (classic formatter combo)
What it is: Black is an opinionated formatter (4 spaces, no config debates)
It can be combined with isort — import sorting, or other tools a typical linter config would need (like flake8)

Pros
 - Very well known (industry standard formatter): easy to explain as a choice
 - Black gives consistent 4-space output
 - Predictable, stable behavior
 - Can take care of converting files that are not in the right format

Cons
 - Multiple tools needed, if more complex formatting or linting is needed (more CI steps, more config)
 - Slightly heavier than Ruff for a greenfield choice today, especially when combined with the additional toolset (isort, flake8)

# Option 4: EditorConfig + manual review (policy without CI gate)
What it is: Add a root .editorconfig or similar file for the most commonly used editors that makes indentation automatic.
This only applies to new edits, and copy-paste, or agent generated code might ignore it.

Editors (VS Code, Cursor, PyCharm) honor it for new edits.

Pros:
 - No runtime dependency
 - Helps developers before commit
 - Very low friction to adopt
 - Can easily deployed together with another option from above

Cons:
 - Not enforced in CI
 - Does not fix existing files
 - Can be overridden by accident

My personal recommendation would be to use option 1, 2 and 4 combined, although it still needs a discussion about what exact level of verification/enforcement is needed.

### Additional Context

_No response_

### Checklist

misi-bp:
# Description

Standardize Python indentation to 4 spaces (PEP 8) for `corto`, `lungo`, and
`recruiter`. Convert remaining tab-indented sources, apply `ruff format`
(Black-compatible layout), and fail CI on tabs, mixed indent, and non-4-space
width.

Ruff is a `dev` extra in each package. A dedicated `Python lint` workflow runs
`ruff check` (not pytest). Editors pick up the same indent from `[tool.ruff]`
and a repo-root `.editorconfig` (`[*.py]` only).

Scout rule: Development agents used not ASCII hyphens that we don't allow. Now the root agents.md makes sure the new content agents generate will keep this rule. 

## Issue Link

Fixes #747

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [x] Other (please describe): contributor tooling — Ruff lint/format, EditorConfig, CI lint workflow

## Checklist


pregnor (commented):
LGTM in general.

I had the agent look at the CI failure of the tests because it was suspicious.

---

LLM review outcome:

## PR 776 scope — not purely Ruff/formatting

The diff is otherwise clean (202 files, almost all `ruff format`/tab→4-space reindentation, plus the new `python-lint.yaml`, `.editorconfig`, `ruff` dev-dependency in the three `pyproject.toml`s, and matching `uv.lock` entries — no existing CI workflow touched). But the **"style(python): use 4-space indentation and apply Ruff formatter"** commit ([db6e999f](https://github.com/agntcy/coffeeAgntcy/commit/db6e999f)) smuggled in two real, non-formatting changes — `ruff format` cannot delete code or change a comparison operand, so these were manual slips during that pass:

1. **Deleted timeout-detection branch** (in both [`corto/exchange/errors.py`](coffeeAGNTCY/coffee_agents/corto/exchange/errors.py) and [`lungo/agents/supervisors/auction/graph/a2a_retry.py`](coffeeAGNTCY/coffee_agents/lungo/agents/supervisors/auction/graph/a2a_retry.py)) — `_is_timeout_error()` lost its `if isinstance(exc, TimeoutError): return True` branch and the matching docstring sentence. **This is what's making `tests / corto` and `tests / lungo` fail in CI** — it's not an environment/SLIM timeout issue at all.
2. **Dead assertion in [`lungo/tests/unit/auction/test_a2a_retry.py`](coffeeAGNTCY/coffee_agents/lungo/tests/unit/auction/test_a2a_retry.py)** — the sleep/backoff check's guard was changed from `if expected_call_count == 3: ... == [1, 2]` to `if expected_call_count == 5: ... == [1, 3, 9, 27]`. No parametrized scenario has `expected_call_count == 5` (max is 3), so this branch is now unreachable — it silently stopped verifying retry-delay timing for the two 3-attempt scenarios (`timeout_then_timeout`, `slim_deadline_exhausted`). It doesn't fail CI itself, just quietly weakens coverage.

Everything else that looked suspicious under a whitespace-ignored diff (`memory.py`, `middleware.py`, `models.py`, `wrapper.py`, `builders.py`, `inflight.py`, `agent_evaluator.py`, `tool_cache_plugin.py`, `version.py`, `docker_helpers.py`, `conftest.py`, etc., across corto/lungo/recruiter) checked out as pure reformatting — verified both by me directly and by the two review agents.

## SLIM timeout CI failure — root cause confirmed

Not a real network/CI timeout. `common/slim_request_timeout.py` (from the just-merged #768) raises a plain `TimeoutError("No SLIM reply from farm within 20s.")`. With regression #1 above, `_is_timeout_error()` no longer recognizes that as a timeout, so:
- `test_is_timeout_error[slim_deadline-True]` fails (`assert False is True`)
- The two retry-then-recover/retry-exhausted scenarios built on it fail too

`recruiter` doesn't have this function at all, which is why its test job passed while corto and lungo failed.

**Fix:** restore the two-line branch in both `errors.py` and `a2a_retry.py`, and fix the `test_a2a_retry.py` guard back to `expected_call_count == 3` / `[1, 2]`. Want me to apply those three fixes now?

pregnor (approved):

mihaialexandrescu (approved):

delthazor (approved):
lgrm

PR #789, opened 2026-09-23
Utilize node type in schema

### LINKED ISSUE #545: [Refactor] Replace Base/AgentExtension node separation with `type` based discriminator
https://github.com/agntcy/coffeeAgntcy/pull/536#discussion_r3182512538

Currently we are using `transportNode` and `customNode` node types for the agentic workflow topology.
On top of this we defined regular, agent and union nodes to handle both the specific and generic treatment of nodes.
We should refactor the code to use different node `type` field values for the agents (like `agent` node type), and base the discriminator of the types on this `type` field to have a cleaner separation.
Hopefully that would simplify the type hierarchy in the JSONSchema as well (we can keep the `AgentNode` == `Node` + `NodeAgentExtension` setup if feasible to have a built up from building blocks experience).

For the implementation we might want to consider transforming the `type` field into a string enum, but it must remain easy to extend.

(also we might want to consider separating the JSON schema file into smaller blocks, but that can be part of another task, breaking the enum - and related types - out into a separate one can be the only thing done here, or doing no separation at all is also fine)

delthazor:
# Description

Node type in schema was present, but was not used to determine exact node type. The presence or absence of other fields were used for that. This change aims to better utilize the type field across all components, and only rely on supplementary fields for backward compatibility.

## Issue Link

Fixes #545 

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [x] Breaking Change
- [x] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Checklist


mihaialexandrescu (dismissed):

mihaialexandrescu (approved):

pregnor (commented):
LGTM, but my agent found some things, evaluate those at your discretion

---

**Correctness (highest priority)**

1. **JSON Schema still presence-only** — the PR only teaches the Python/Pydantic discriminator to trust `type`; `event_v1.json`'s `node_type` def has no `enum`/`if-then`, so a payload can be schema-valid but Pydantic-invalid. Four places (JSON Schema, Pydantic discriminator, `make_node`, frontend) now each encode a slightly different classification rule.
2. **`make_node()` can raise `pydantic.ValidationError`** for `type=agent/mcp` nodes with no `stable_agent_id`/`agent_record_uri` — dormant today (every real caller passes `stable_agent_id`), but reachable through `middleware.py`'s uncaught interceptor on real agent-to-agent traffic if a future caller regresses.
3. **Uncaught `ValidationError` in `instantiate_agentic_workflow`** — the endpoint only catches `SchemaValidationError`, not `pydantic.ValidationError`, so a future catalog entry with `type=agent` but no extension (still schema-valid by design) would 500 instead of returning a clean error.
4. **`make_node` never checks `BASE_NODE_TYPES`** — can classify a `directory` node as agent-shaped while the wire discriminator would classify the same payload as base.
5. **Discriminator checks `BASE_NODE_TYPES` before presence** — a mistagged `transport`/`group`/`directory` node carrying stray agent fields silently lands as an untyped extra instead of failing validation.
6. **Frontend icon regression** — `topologyNodeIconKind`'s new `isMcpType` branch checks only the label substring and skips the pre-existing slug-based fallback, so relabeled MCP nodes could lose their correct icon.

**Altitude / design**

7. **New type checks are OR'd with old label heuristics, not replacing them** — `isDirectoryType(nodeType) || isDirectoryLabel(...)` etc. across 4+ call sites means label-guessing can still override the authoritative `type` field, reintroducing the exact issue #545 problem one layer up.

**Simplification / maintainability**

8. **`NODE_TYPES` vs `NODE_TYPE`** in `const.ts` — near-identical names, overlapping values, easy to grab the wrong one.
9. **Dead code, pre-existing** — the new `isGroupType` OR-check in `enrichAgenticTopologyWellKnownUi` is unreachable in production (intercepted earlier in the pipeline); the PR touches the line without fixing it.
10. **Duplicated weather/payment icon-matching logic**, verbatim, in two functions.
11. **Discriminator logic hand-copied into docs** (`mapping-rules.md`/`SKILL.md`) with no automated check that it still matches the code.

**Efficiency**

12. **`canonicalTypeOf`/`rfIdOf` recomputed ~7-10x per node** in `topologyWireToReactFlow`, on a live re-render hot path — not a scaling bug, but wasted CPU worth collapsing into one pass.

delthazor:
> LGTM, but my agent found some things, evaluate those at your discretion
> 
> **Correctness (highest priority)**
> 
> 1. **JSON Schema still presence-only** — the PR only teaches the Python/Pydantic discriminator to trust `type`; `event_v1.json`'s `node_type` def has no `enum`/`if-then`, so a payload can be schema-valid but Pydantic-invalid. Four places (JSON Schema, Pydantic discriminator, `make_node`, frontend) now each encode a slightly different classification rule.
> 2. **`make_node()` can raise `pydantic.ValidationError`** for `type=agent/mcp` nodes with no `stable_agent_id`/`agent_record_uri` — dormant today (every real caller passes `stable_agent_id`), but reachable through `middleware.py`'s uncaught interceptor on real agent-to-agent traffic if a future caller regresses.
> 3. **Uncaught `ValidationError` in `instantiate_agentic_workflow`** — the endpoint only catches `SchemaValidationError`, not `pydantic.ValidationError`, so a future catalog entry with `type=agent` but no extension (still schema-valid by design) would 500 instead of returning a clean error.
> 4. **`make_node` never checks `BASE_NODE_TYPES`** — can classify a `directory` node as agent-shaped while the wire discriminator would classify the same payload as base.
> 5. **Discriminator checks `BASE_NODE_TYPES` before presence** — a mistagged `transport`/`group`/`directory` node carrying stray agent fields silently lands as an untyped extra instead of failing validation.
> 6. **Frontend icon regression** — `topologyNodeIconKind`'s new `isMcpType` branch checks only the label substring and skips the pre-existing slug-based fallback, so relabeled MCP nodes could lose their correct icon.
> 
> **Altitude / design**
> 
> 7. **New type checks are OR'd with old label heuristics, not replacing them** — `isDirectoryType(nodeType) || isDirectoryLabel(...)` etc. across 4+ call sites means label-guessing can still override the authoritative `type` field, reintroducing the exact issue #545 problem one layer up.
> 
> **Simplification / maintainability**
> 
> 8. **`NODE_TYPES` vs `NODE_TYPE`** in `const.ts` — near-identical names, overlapping values, easy to grab the wrong one.
> 9. **Dead code, pre-existing** — the new `isGroupType` OR-check in `enrichAgenticTopologyWellKnownUi` is unreachable in production (intercepted earlier in the pipeline); the PR touches the line without fixing it.
> 10. **Duplicated weather/payment icon-matching logic**, verbatim, in two functions.
> 11. **Discriminator logic hand-copied into docs** (`mapping-rules.md`/`SKILL.md`) with no automated check that it still matches the code.
> 
> **Efficiency**
> 
> 12. **`canonicalTypeOf`/`rfIdOf` recomputed ~7-10x per node** in `topologyWireToReactFlow`, on a live re-render hot path — not a scaling bug, but wasted CPU worth collapsing into one pass.

We already have a follow-up ticket that is aimed to take care of several of these issues. I added it to the description.

1: Intentional, target of the follow-up ticket
2: Intentional, has a proper test coverage, and unlikely in live run
3: fixed
4: fixed
5: intentional, will be cleaned with the follow-up ticket
6: fixed
7: Intentional, target of the follow-up ticket
8: fixed
9: Needed for backward compatibility on legacy CustomNodes (not entirely dead code). I will inevitably be removed with the implementation of the follow-up ticket
10: fixed
11: intentional, target of the follow-up ticket
12: fixed

pregnor (approved):

delthazor:
Run the app locally, to make a manual UI smoke test, and it seems fine for the first glance, no visible regression.

mihaialexandrescu (approved):

